PR #5, opened 2019-03-11
Feature/transit stops

mikkoisa:
Add transit stops with timetables to high zoom-level

* Created two different fetch functions, one for transit stops and another for subway entrances
* Combined subway entrance data with stops data and added arrival schedule to entrance info
* Removed subway stations from data and replaced them with entrances

### DIFF
--- a/src/views/Map/components/MapView.js
+++ b/src/views/Map/components/MapView.js
@@ -51,7 +51,7 @@ class MapView extends React.Component {
             maxZoom={mapBase.options.maxZoom}
             maxBounds={mapOptions.maxBounds}
             onMoveEnd={() => {
-              if (this.mapRef.current.leafletElement._zoom >= 14) {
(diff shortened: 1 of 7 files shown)


src/views/Map/components/MapView.js
```diff
@@ -1,51 +1,62 @@
...
-            // id="mapid"
             zoomControl={false}
             crs={mapBase.crs}
             center={mapOptions.initialPosition}
             zoom={mapBase.options.zoom}
             minZoom={mapBase.options.minZoom}
             maxZoom={mapBase.options.maxZoom}
             maxBounds={mapOptions.maxBounds}
+            onMoveEnd={() => {
+              if (this.mapRef.current.leafletElement._zoom >= 14) {
```
Dremars: Transit stop zoom level should be set in map options

PR #10, opened 2019-03-20
Feature/map districts

mikkoisa:
Added district fetch and drawing on map functionality, with mock location for now (Kluuvi)
Restructured folder structure: moved Map-Specific utils and constants into Map-folder


src/views/Map/MapContainer.js
```diff
@@ -96,20 +122,23 @@ class MapContainer extends React.Component {
...
           mapOptions={mapOptions}
           fetchTransitStops={this.fetchTransitStops}
           clearTransitStops={this.clearTransitStops}
           transitStops={transitStops}
           t={id => translate(state, id)}
           // TODO: think about better styling location for map
-          style={{ width: '100%', height: '92.6%', position: 'absolute' }}
+          style={{
+            width: '100%', height: '92.6%', position: 'absolute', zIndex: -1,
+          }}
```
Dremars: Inline styles need to be changed to use material UI (cssInJS) withStyles method.

mikkoisa: Styling will be changed in another PR with all other inline stylings

PR #19, opened 2019-04-01
Feature/unit view client side fetch

mikkoisa:
Execute fetch on client-side, that fetches all of the unit details and overrides te old unit data from server-side


src/utils/queryBuilder.js
```diff
@@ -13,6 +13,13 @@ const searchQueryData = {
   geometry: true,
 };
 
+const unitQueryData = {
```
Dremars: Are these query parameters used for anything? Unit query is already specified to get certain unit with ID /unit/{unitID}/  So page, page_size are unnecessary and I'm not sure if include and geometry are used?

src/views/UnitView/UnitView.js
```diff
@@ -6,31 +6,45 @@ import { FormattedMessage } from 'react-intl';
...
-      console.log('change selected unit to: ', unit);
       changeSelectedUnit(unit);
+      console.log('change selected unit to: ', unit);
+
+      this.setState({ isFetching: true });
+      queryBuilder.setType('unit', unit).run()
+        .then(response => response.json())
+        .then(data => this.setState({
+          unitData: data,
+          isFetching: false,
```
Dremars: Unit data should be fetched with redux action so state stays consistent. IsFetching also should come from redux so it doesn't mutate unexpectedly.

mikkoisa: Will be fixed soon

PR #24, opened 2019-04-04
Feature/path generation

Dremars:
Adding paths to own file and creating function to generate path strings


Dremars:
I'm not sure if `/config` folder is the best place for this. Maybe this should be in utils?

PR #26, opened 2019-04-08
Fix/unit view styling

mikkoisa:
* Added material ui typography to themes
* Improved unit-view styling
* Added Internationalization to unit-view fields
* Added locale to redux on first server side render.


src/views/UnitView/UnitView.js
```diff
@@ -107,10 +117,12 @@ class UnitView extends React.Component {
               title={<FormattedMessage id="unit.contact.info" />}
             />
             <InfoList // E-services
+              getLocaleText={getLocaleText}
```
Dremars: This selector could be connected straight to InfoList

PR #29, opened 2019-04-11
Feature/unit view improved fetch

mikkoisa:


src/redux/reducers/unit.js
```diff
@@ -31,6 +31,19 @@ export default (state = initialState, action) => {
         count: 0,
         max: 0,
       };
+    case 'UNITS_UPDATE_FETCH_SUCCESS':
+      return {
+        ...state,
+        isFetching: false,
+        errorMessage: null,
+        // This checks which unit on the list was updated and updates its data
+        data: state.data.map((unit) => {
```
Dremars: If unit doesn't exist in current state.data unit could be added to new data

PR #41, opened 2019-04-17
Fix service page screen reader focus

mikkoisa:


Dremars:
On first time entering service page focus and alert works but on second time it doesn't focus or say anything

PR #43, opened 2019-04-23
Feature/center to unit marker

mikkoisa:
Map zooms to unit marker location when entering unit details page
* Add reference to map to redux, so map actions can be called anywhere
* Added functionality to automatically resize map to fit all search results to view. Not used currently


src/views/UnitView/UnitView.js
```diff
@@ -87,6 +85,12 @@ class UnitView extends React.Component {
     }
 
     if (unit && unit.complete) {
+      if (map) {
```
Dremars: This could break once the view rerenders. For example when we get tabs for UnitView the view might rerender and refocus map. Is it possible to set in componentdidmount?

Dremars:
Refreshing UnitView breaks script

PR #53, opened 2019-04-29
Add tests, small fixes

mikkoisa:
* Add browser tests for unitView
* Add automatic axe tests to all pages


src/views/Map/components/MapView.js
```diff
@@ -152,6 +152,7 @@ class MapView extends React.Component {
             if (unit && unit.location) {
               return (
                 <Marker
+                  id="unitMarker"
```
Dremars: Should be a class since this mapping draws all unit markers based on filters. Tests can use selector to look for class name.

PR #55, opened 2019-05-02
Add example searches to front page

mikkoisa:


Dremars:
* I think Container changes are breaking Titlebar padding. Can be observed in unit view.
* Also divider on frontpage could be removed.

PR #67, opened 2019-05-08
Fix/mobile breakpoint

mikkoisa:
* Fix issue with mobile layout not appearing on mobile devices
* Fix issues in mobile view caused by addressbar appearing/disappearing 
* Made map refresh when opened  on mobile so that the tiles load properly

### DIFF
--- a/src/views/Map/MapContainer.js
+++ b/src/views/Map/MapContainer.js
@@ -146,7 +146,7 @@ class MapContainer extends React.Component {
           transitStops={transitStops}
           getLocaleText={textObject => getLocaleText(textObject)}
           // TODO: think about better styling location for map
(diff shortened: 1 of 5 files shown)


src/views/Map/MapContainer.js
```diff
@@ -199,6 +199,7 @@ MapContainer.propTypes = {
   fetchDistrictsData: PropTypes.func.isRequired,
   getLocaleText: PropTypes.func.isRequired,
   setMapRef: PropTypes.func.isRequired,
+  // isMobile: PropTypes.bool,
```
Dremars: These comments could be removed if there's no plan to use these in imminent future

PR #71, opened 2019-05-10
Fix feedback text alignment

Dremars:
* Fix Navbar text alignment to align with other items

### DIFF
--- a/src/layouts/DefaultLayout.js
+++ b/src/layouts/DefaultLayout.js
@@ -3,7 +3,7 @@ import React from 'react';
 import { withRouter } from 'react-router-dom';
 import PropTypes from 'prop-types';
 import {
-  Button, Grid, Typography, AppBar, Toolbar,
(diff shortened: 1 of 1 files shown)


Dremars:
Noticed there's a problem when resizing window. At lower widths before mobile view feedback text goes to multiple lines and pushes other elements up to top of the navbar edge.
If you have any ideas how to fix this feel free to do changes.

mikkoisa:
Noticed the same issue while adding the feedback link. Decided to uncenter the text intentionally to prevent that happening, because the issue can be seen clearly on tablet devices. Having all three elements inside the same grid doesn't seem to work, so one solution could be to take the feedback text out of the grid and position it on the top of it.

Dremars:
[Web-yläpalkin elementtien vertikaalinen tasaus](https://trello.com/c/hqKEaaB2/107-web-yl%C3%A4palkin-elementtien-vertikaalinen-tasaus)

PR #76, opened 2019-05-15
Move jump link to first element on page

Dremars:


Dremars:
[Hyppylinkki aina sivun ensimmäiseksi elementiksi](https://trello.com/c/arQJM90w/126-hyppylinkki-aina-sivun-ensimm%C3%A4iseksi-elementiksi)

mikkoisa:
Focus when first loading the page is on the icon/title, which is after the jump link. Doesn't this make the jump link hard to reach, since it is before the entry point to the page? The focus should probably be earlier

PR #82, opened 2019-05-17
Feature/unit view mobile version

mikkoisa:
* Modify title bar (used in unit-view and service-view) into different versions for mobile and desktop
* Use helmet to change the title tag of each page. (Should discuss what these titles should be)
* Use helmet to change the status bar and address bar color to blue on unit-view and service-view. (Experimental, should discuss do we want it to behave like this)


mikkoisa:
[Sivun titlen dynaaminen päivittäminen](https://trello.com/c/AbS6Hjg0/132-sivun-titlen-dynaaminen-p%C3%A4ivitt%C3%A4minen)

src/views/HomeView/HomeView.js
```diff
@@ -39,10 +40,22 @@ class HomeView extends React.Component {
     }
   }
 
+
   render() {
     const { intl, classes } = this.props;
+
+    // Modify html head
+    const Head = (
```
Dremars: I think we should create one component to handle Helmet attributes and use it only in one place (layout file). This way we can keep code cleaner and in one place. We could utilize router's location parameter and path regex to check current page in order to figure out title text.

src/views/UnitView/UnitView.js
```diff
@@ -85,17 +96,33 @@ class UnitView extends React.Component {
...
+      </Helmet>
+    );
 
     const TopBar = (
       <div>
-        <TitleBar title={unit && unit.name ? getLocaleText(unit.name) : ''} />
+        <DesktopComponent>
+          <SearchBar placeholder={intl.formatMessage({ id: 'search' })} />
+        </DesktopComponent>
+        <TitleBar icon={icon} title={unit && unit.name ? getLocaleText(unit.name) : ''} />
```
Dremars: This could be calculated at the start of the render function and assigned to a variable to keep the components cleaner and more readable.

src/views/components/HeadInfo.js
```diff
@@ -0,0 +1,63 @@
...
+  const message = messageId ? intl.formatMessage({ id: messageId }) : '';
+  let pageMessage = '';
+
+  // Add unit or service name to title if needed
+  if (page === 'unit' && unit && unit.name) {
+    pageMessage = getLocaleText(unit.name);
+  } if (page === 'service' && service && service.name) {
+    pageMessage = getLocaleText(service.name);
+  }
+  const title = `${message} ${uppercaseFirst(pageMessage)}`;
```
Dremars: Title could be for example "Kallion kirjasto | Palvelukartta" or "Kallion kirjasto - Palvelukartta"  format.

PR #83, opened 2019-05-17
Fix/mobile usability

mikkoisa:
* Fix issue with list-item reading with NVDA
* Make default zoom on map smaller on mobile to show larger area.
* Show transit stops earlier (on lower zoom) on mobile version
* Fetch fewer transit stops to improve map performance on mobile devices
* Fix issue with footer back-to-start-link not working on mobile. (was trying to focus on site-title which does not exist on mobile)
* Move sr-only title on search page out of window, because it made clicking some elements more difficult on mobile screen reader


mikkoisa:
[Palaa sivun alkuun –ankkurilinkki ei toimi iPhonen VoiceOverilla](https://trello.com/c/d1RtaOHd/131-palaa-sivun-alkuun-ankkurilinkki-ei-toimi-iphonen-voiceoverilla)

mikkoisa:
[Linkilliset yhteystietotiedot koodataan, kuten kotisivutieto](https://trello.com/c/QLyMjkHv/127-linkilliset-yhteystietotiedot-koodataan-kuten-kotisivutieto)

src/views/Map/components/MapView.js
```diff
@@ -88,8 +88,9 @@ class MapView extends React.Component {
 
   // Check if transit stops should be shown
   showTransitStops() {
-    const { mapBase } = this.props;
-    return this.mapRef.current.leafletElement._zoom >= mapBase.options.transitZoom;
+    const { mapType, mobile } = this.props;
+    const transitZoom = !mobile ? mapType.options.transitZoom : mapType.options.mobileTransitZoom;
+    return this.mapRef.current.leafletElement._zoom >= transitZoom;
```
erikjuhani: Explain it better. What does a truthy value imply?

The underscore in front of the `_zoom` attribute usually indicates that it is meant to be a private attribute of the class. You should rather access it via a _public function_ 
```js
this.getZoom();
```

https://leafletjs.com/reference-1.4.0.html#map-getzoom

Besides this whole return statement seems a bit complex. Perhaps making it in several parts would make it more clear.

* Create a function to return the leaflet element
* Split it up
```js
const transitZoom = ... ;
const mapRefElement = this.getMapRefElement();
const currentZoom = mapRefElement.getZoom();

return zoom >= transitZoom;
```

Just my two 💵s

src/views/Map/components/MapView.js
```diff
@@ -134,11 +136,11 @@ class MapView extends React.Component {
           keyboard={false}
           style={style}
           zoomControl={false}
-          crs={mapBase.crs}
+          crs={mapType.crs}
           center={mapOptions.initialPosition}
-          zoom={mapBase.options.zoom}
-          minZoom={mapBase.options.minZoom}
-          maxZoom={mapBase.options.maxZoom}
+          zoom={!mobile ? mapType.options.zoom : mapType.options.mobileZoom}
```
erikjuhani: Separating this ternary to it's own variable makes sense here (to increase readability in my opinion).
After this separation you're left with a simple zoom variable.

src/views/Map/utils/transitFetch.js
```diff
@@ -7,11 +7,11 @@ const fetchStops = async (bounds) => {
...
-  // Widen the area of search to be three times the size of the view
-  fetchBounds._southWest.lat -= viewSize.height * 2;
-  fetchBounds._southWest.lng -= viewSize.width * 2;
-  fetchBounds._northEast.lat += viewSize.height * 2;
-  fetchBounds._northEast.lng += viewSize.width * 2;
+  // Widen the area of search
+  fetchBounds._southWest.lat -= viewSize.height;
+  fetchBounds._southWest.lng -= viewSize.width;
+  fetchBounds._northEast.lat += viewSize.height;
+  fetchBounds._northEast.lng += viewSize.width;
```
erikjuhani: As an _outsider_ this comment makes me have to read the actual code of what it does 🙂

More explicit comment would be something like this:
`// Increase the area of search by the amount of current view size`

Ok I'm sorry for repeating myself, but seems quite _dangerous_ that you are actually manipulating a _"private"_ attribute of a class and setting a new value to it... Instead you should probably make a new `LatLngBounds` object with the new calculated bound corners.

```js
const corner1 = ... ;
const corner2 = ... ;
const wideBounds = L.latLngBounds(corner1, corner2);
```
https://leafletjs.com/reference-1.4.0.html#latlngbounds-l-latlngbounds

https://github.com/City-of-Helsinki/servicemap-ui/pull/83/files#diff-4312092f55ff21a9da5ece981aac368dR2
Another note, the actual parameter this function receives is not `bounds`, but the `leafletElement`. I would suggest to change it.

Just my two 🔦s

src/layouts/DefaultLayout.js
```diff
@@ -199,14 +200,14 @@ const DefaultLayout = (props) => {
               label: intl.formatMessage({ id: 'map' }),
               onClick: () => {
                 if (history) {
-                  // TODO: Add query text once functionality is ready for search view
+                // TODO: Add query text once functionality is ready for search view
```
Dremars: These TODOs are remnants of certain history pushes related to search view. Must have been copied here from somewhere. Should be removed since they are misleading.

src/layouts/DefaultLayout.js
```diff
@@ -188,7 +189,7 @@ const DefaultLayout = (props) => {
               label: intl.formatMessage({ id: 'general.home' }),
               onClick: () => {
                 if (history) {
-                  // TODO: Add query text once functionality is ready for search view
+                // TODO: Add query text once functionality is ready for search view
```
Dremars: These TODOs are remnants of certain history pushes related to search view. Must have been copied here from somewhere. Should be removed since they are misleading.

PR #85, opened 2019-05-22
Fix/refactor component locations

Dremars:
Small fixes and refactoring
* Move pagination component to components folder
* Split pagination component's styles to own file
* Add index file for pagination component
* Remove unused redux store
* Calculate showed elements before returning render

### DIFF
--- a/src/components/TabLists/PaginationComponent/styles.js
+++ b/src/components/TabLists/PaginationComponent/styles.js
@@ -1,0 +1,14 @@
+
+// Styles
+export default theme => ({
+  button: {
+    padding: `${theme.spacing.unit / 2}px ${theme.spacing.unit}px`,
+    margin: theme.spacing.unit / 2,
+    minWidth: 'auto',
+  },
+  pageItem: {
+    margin: 0,
+    marginLeft: theme.spacing.unit / 2,
+    marginRight: theme.spacing.unit / 2,
+  },
+});
--- a/src/components/TabLists/PaginationComponent/styles.js
+++ b/src/components/TabLists/PaginationComponent/styles.js
@@ -2,13 +2,13 @@
 // Styles
 export default theme => ({
   button: {
-    padding: `${theme.spacing.unit / 2}px ${theme.spacing.unit}px`,
-    margin: theme.spacing.unit / 2,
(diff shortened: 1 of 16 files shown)


src/views/SearchView/SearchView.js
```diff
@@ -198,16 +201,13 @@ class SearchView extends React.Component {
...
           && (
             <TabLists data={searchResults} sortCallback={this.sortCallback} />
           )
         }
         {
-          !isFetching
-          && previousSearch
-          && units
-          && units.length === 0
+          showNotFoundText
```
erikjuhani: I personally would separate this logic into it's own function. To keep this in smaller logical parts. It's usually quite exhausting to read long functions (this one is very long). 
```jsx
// Here
{
  this.renderNotFound()
}

// Elsewhere
renderNotFound = () => {
  const { units } = this.props;

  // These variables should be passed to this function
  const shouldRender = !isFetching && previousSearch && units && !units.length;
  
  return shouldRender && (
    <Container>
      <Typography variant="subtitle1" component="p" aria-hidden="true">
        <FormattedMessage id="search.results" values={{ count: units.length }} />
      </Typography>
    </Container>
  );
}
```

src/components/TabLists/PaginationComponent/styles.js
```diff
@@ -0,0 +1,14 @@
...
+export default theme => ({
+  button: {
+    padding: `${theme.spacing.unit / 2}px ${theme.spacing.unit}px`,
+    margin: theme.spacing.unit / 2,
+    minWidth: 'auto',
+  },
+  pageItem: {
+    margin: 0,
+    marginLeft: theme.spacing.unit / 2,
+    marginRight: theme.spacing.unit / 2,
```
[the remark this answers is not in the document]
Dremars: True and thanks for the comment. I think having it once in theme would probably be the best. I'll do some changes.

src/views/SearchView/SearchView.js
Dremars: Good point. And I would agree those bigger components are getting out of hands. This method would make it much easier to read and understand. We'll start using this from now on. Thank you!

PR #87, opened 2019-05-23
Feature/dynamic back button label

Dremars:
**New dynamic back button texts**
New text is calculated from current location and then translated using react-intl's translation functionality.
Attempts to change text and aria-label of button to  `Back to home view` etc. and defaults to `Back` if no text with given path is found in translations file.

* Change breadcrumb to redux and update Navigator component
* Calculate dynamic back button text based on current location in app

### DIFF
--- a/src/components/BackButton/BackButton.js
+++ b/src/components/BackButton/BackButton.js
@@ -13,7 +13,7 @@ const BackButton = (props) => {
 
   // Generate dynamic text
   // Figure out correct translation id suffix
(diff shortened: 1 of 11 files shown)


Dremars:
[Kerrotaan dynaamisesti minne paluulinkki ohjaa](https://trello.com/c/mp6WxWWD/130-kerrotaan-dynaamisesti-minne-paluulinkki-ohjaa)

PR #90, opened 2019-06-05
Feature/unit view events

mikkoisa:
* Add events to redux and unit view
* Break unitView to components (description, contact info, highlights, events, services, e-services...)
* Add new pages: 
EventDetailView - displays the event's information when event is clicked from the list  
UnitEventsView - the view that appears when "show all events" is clicked.
* Add new reusable components:
BlueButton - the button for "show more events/services",  "send feedback" etc.
DescriptionText - the text used in unit description and event description
* Modify server unitHandler to work also with the new /events url


mikkoisa:
TODO: add correct html titles to new pages, fix missing icon in event details title

mikkoisa:
[Tapahtuman oma sivu](https://trello.com/c/6luUfLb8/36-tapahtuman-oma-sivu)

src/components/BlueButton/BlueButton.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Button, withStyles } from '@material-ui/core';
+
+
+const BlueButton = ({
```
Dremars: Button wrapper component is good idea but we should change the name to something more generic. This button isn't blue if primary color is changed. In the future we might also want to change button's variant etc. but those changes can be made once it's necessary. For now we just need more generic name. It  could also be just <SMButton /> for Service Map button. And use this naming scheme in the future when extending MUI components.

PR #92, opened 2019-06-11
Add show on map button to mobile unit view

mikkoisa:
* Add show on map button to mobile unit view
* Add top bar to map (is either search bar or title bar depending on page)
* Modify mobile bottom navigation so that the correct page is highlighted when coming back to homepage using back-button.
* Add new variable fullMobileMap to defaultLayout styling. If fullMobileMap is true, use styling for map view without bottom navigation and search-bar

### DIFF
--- a/config/paths.js
+++ b/config/paths.js
@@ -13,12 +13,8 @@ const paths = {
     regex: /\/[a-zA-Z]{2}\/search/
   },
   unit: {
-    generate: id => `/unit/${id || ''}`,
+    generate: data => `/unit/${data.id || ''}${data.query || ''}`,
     regex: /\/[a-zA-Z]{2}\/unit\/([0-9]+)/
-  },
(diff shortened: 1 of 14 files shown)


Dremars:
Managed to get this weird white bar at the bottom of map. Doesn't seem to happen on Android.
Tested with Chrome Version 68.0.3440.75 and Firefox 66.0.2

sm_mobile_whitebar

mikkoisa:
When shrinking the browser window, the app thinks that it is on landscape mode (landscape mode is defined by device aspect ratio), so it applies the wrong styles.

mikkoisa:
[Uusi media-query toteutus laitteen tunnistamiseksi](https://trello.com/c/UTZ0XFj0/158-uusi-media-query-toteutus-laitteen-tunnistamiseksi)

PR #94, opened 2019-06-13
Add different list for school unit services

mikkoisa:
On school units, list services with periodical data differently than normal services.

### DIFF
--- a/src/views/UnitView/components/Services.js
+++ b/src/views/UnitView/components/Services.js
@@ -1,24 +1,117 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { FormattedMessage } from 'react-intl';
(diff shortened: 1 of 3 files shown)


src/views/UnitView/components/Services.js
```diff
@@ -1,24 +1,117 @@
...
+    this.state = {
+      serviceList: [],
+      subjectList: [],
+      periodList: [],
+    };
+  }
+
+  componentDidMount() {
+    const { unit } = this.props;
+    if (unit.root_service_nodes.includes(1087)) {
```
Dremars: Lets move this parameter to a variable in either start of the file, or constants file so it's easily modifiable if necessary.

PR #98, opened 2019-06-19
Add area to map on units with geometry data

mikkoisa:
* Add geometry data to unit fetches
* Remove useless fields from unit search fetch to make search faster
* On mobile, make map hidden instead of display: none, so that map actions work correctly
* Move some of the map functions from MapContainer to MapView (the Map components will be refactored further soon)
* Add area drawing to map on units with geometry data (skiing tracks, running paths etc.)


mikkoisa:
[Kartalle näytettävät tiedot (ladut, piirit ym.)](https://trello.com/c/dJMzP1OB/41-kartalle-n%C3%A4ytett%C3%A4v%C3%A4t-tiedot-ladut-piirit-ym)

PR #103, opened 2019-06-26
Feature/mobile top bar

mikkoisa:
* Create new topBar component which handles desktop and mobile topBars
* Move settings component to sidebarWrapper. Move settings button from settings component to topBar component
* Move some functionality from settings component to defaultLayout (open state)
* Make mobile friendly version of settings component
* Add new temporary icon for mobile topBar


src/components/Logos/HomeLogo/index.js
```diff
@@ -2,13 +2,21 @@ import React from 'react';
...
+  const { dark, mobile, ...rest } = props;
   if (dark) {
     return (
       <div role="img" {...rest}>
-        <img src={logoDark} alt="" style={{ height: 36 }} />
+        <img src={logoDark} alt="" style={{ marginTop: 12, marginBottom: 12, height: 45 }} />
+      </div>
+    );
+  }
+  if (mobile) {
```
Dremars: Mobile check should be before dark. Otherwise if dark is set to true while on mobile logo will not be mobile logo.

src/views/Map/MapContainer.js
```diff
@@ -56,19 +56,26 @@ class MapContainer extends React.Component {
...
-      <div style={{
-        zIndex: 99999999, position: 'fixed', top: 0, width: '100%',
-      }}
-      >
+      <>
         {isMobile && currentPage === 'map' && (
           // If on root map page (/map) display search bar.
-          <SearchBar placeholder={intl.formatMessage({ id: 'search' })} />
+          <div style={{
+            zIndex: 99999999, position: 'fixed', top: 64, width: '100%',
```
Dremars: It seems like this top: 64 is used in multiple places. Isn't this related to topbar's height? If so I think this could be set in constants or somewhere where it can be changed once and affects all places it's used.

mikkoisa: https://trello.com/c/F4vRUzMs/174-topbar-korkeus-globaaliksi-arvoksi

PR #104, opened 2019-06-27
Fetch and use selected unit's data in event defails

Dremars:
Changed EventDetailView to use selectedUnit rather than event's location. Data seems mostly correct but telephone is actually phone in data coming from ServiceMap API.

### DIFF
--- a/src/views/EventDetailView/EventDetailView.js
+++ b/src/views/EventDetailView/EventDetailView.js
@@ -77,7 +77,7 @@ class EventDetailView extends React.Component {
 
   render() {
     const {
-      event, intl, getLocaleText, navigator, selectedUnit,
+      event, intl, getLocaleText, selectedUnit,
     } = this.props;
     if (event) {
       const description = event.description || event.short_description;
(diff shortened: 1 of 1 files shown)


src/views/EventDetailView/EventDetailView.js
```diff
@@ -102,14 +122,19 @@ class EventDetailView extends React.Component {
...
-            />
+            {
+               phoneText
+               && (
+               <SimpleListItem
+                 key="contactNumber"
+                 icon={<Phone />}
+                 text={phoneText}
+                 srText={intl.formatMessage({ id: 'unit.phone' })}
+                 link
```
Dremars: By the way, I noticed that phone item is set as link even though it doesn't seem to do anything or take user to anywhere. Is this missing onClick action or is link unnecessary?

mikkoisa: @Dremars The link will be used to directly make phone call. Just hasn't been implemented yet.

PR #111, opened 2019-07-05
Check if divider needed on unitView lists

mikkoisa:
+ Add check to unitView lists that define if divider is needed (last item on the list should not display divider)
+ Fix issue with units with null contact info displaying empty list
+ Refactor school period service listing


mikkoisa:
[Palveluja listattaessa viimeisellä palvelulla ei saa olla harmaata div-viivaa](https://trello.com/c/vXKIw7KG/177-palveluja-listattaessa-viimeisell%C3%A4-palvelulla-ei-saa-olla-harmaata-div-viivaa)

PR #114, opened 2019-07-09
Feature/unit view tabs

Dremars:
- Update Search- and Titlebar stylings
- Refactored TitleBar
- Updated TabLists heading elements and tabs to be sticky

### DIFF
--- a/src/components/Container/Container.js
+++ b/src/components/Container/Container.js
@@ -22,6 +22,13 @@ const styles = theme => ({
   text: {
     textAlign: 'left',
   },
+  noMargin: {
+    margin: 0,
+    paddingLeft: theme.spacing.unit,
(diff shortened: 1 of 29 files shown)


Dremars:
[Toimipisteen tietojen esittäminen välilehdillä](https://trello.com/c/pLJZ2O7E/153-toimipisteen-tietojen-esitt%C3%A4minen-v%C3%A4lilehdill%C3%A4)

mikkoisa:
Also noticed that since the accessibility info tab content doesn't have a title, the first shortcomings are h5 (h4 is skipped). 

Näyttökuva 2019-7-11 kello 10 11 28


Also, should we consider making the tab-buttons headings as well (h4)? So that people who navigate by headings won't miss content.

PR #126, opened 2019-08-06
Add TopArea component to all pages

mikkoisa:
Added blue color top area to each page:
* Created new component TopArea, which handles the searchbar/titlebar area of each page.
* Changed service berry drawing colors to white on dark blue backgrounds

### DIFF
--- a/src/views/MapView/utils/drawIcon.js
+++ b/src/views/MapView/utils/drawIcon.js
@@ -144,25 +144,26 @@ export const drawUnitIcon = (berryColor, curve) => {
 };
 
 // Temporary solution
-export const drawServiceIcon = () => {
+export const drawServiceIcon = (primary) => {
   const canvas = document.createElement('canvas');
   const ctx = canvas.getContext('2d');
   canvas.height = canvasSize.height;
   canvas.width = canvasSize.width;
   const berryColor = 'gray';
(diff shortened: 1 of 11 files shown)


src/views/MapView/utils/drawIcon.js
```diff
@@ -144,25 +144,26 @@ export const drawUnitIcon = (berryColor, curve) => {
 };
 
 // Temporary solution
-export const drawServiceIcon = () => {
+export const drawServiceIcon = (primary) => {
```
Dremars: We should use white borders for all service icons which will make figuring out primary irrelevant. Check unit berry drawings for reference

src/views/ServiceView/ServiceView.js
```diff
@@ -31,7 +29,7 @@ class ServiceView extends React.Component {
     const { params } = match;
 
     this.setState({
-      icon: <img alt="" src={drawServiceIcon()} style={{ height: 24 }} />,
+      icon: <img alt="" src={drawServiceIcon('primary')} style={{ height: 24 }} />,
```
Dremars: Primary can be removed if border is set for all icons.

Dremars:
@mikkoisa This seems old and deprecated. Can we close this?

PR #142, opened 2019-09-19
Fix/city filtering

Dremars:
- Add city filtering to search result selector
- Add city filtering to service units selector
- Change map to use selectors to get results and services
- Remove old city filtering in ServiceView


Dremars:
[Piilotetaan tyhjät palvelut](https://trello.com/c/AbtbYWdk/124-piilotetaan-tyhj%C3%A4t-palvelut)

PR #144, opened 2019-09-19
Add new no results text to search view

Dremars:
- New text for no results in SearchView
- Hide Expand search button if result list is empty


Dremars:
Need to check that accessibility is ok.

PR #148, opened 2019-09-25
Fix/ssr for unit and event view

Dremars:
- Added server side data fetching required by SSR functionality.
- Moved accessibility sentences and reservations to redux state
- Fixed problems in components caused by SSR data fetching


Dremars:
Noticed problem when loading page while on event view. Description text is left empty.

Dremars:
[Server-side fetchaus kaikille sivuille](https://trello.com/c/d4D5oegC/186-server-side-fetchaus-kaikille-sivuille)

PR #164, opened 2019-10-21
Add aria label text for map element

Dremars:
Adding aria-label to map wrapper to inform user that map exists. This way map settings in settings page are not as confusing for screen reader users.


Dremars:
[Kartalle asetetaan aria-label joka kertoo kyseessä olevan kartta](https://trello.com/c/FzevD07E/251-kartalle-asetetaan-aria-label-joka-kertoo-kyseess%C3%A4-olevan-kartta)

PR #167, opened 2019-10-25
Add nearby services button to home view

Dremars:


Dremars:
[Lähellä olevat palvelut](https://trello.com/c/7tOMnm3m/51-l%C3%A4hell%C3%A4-olevat-palvelut)

mikkoisa:
The button could be disabled if no user location found (or somehow indicate that it doesn't work). This could be done when the new front-page is implemented.

PR #168, opened 2019-10-28
Add sorting based on distance

Dremars:


Dremars:
[Hakutulosten järjestäminen etäisyyden mukaan](https://trello.com/c/InidhObr/254-hakutulosten-j%C3%A4rjest%C3%A4minen-et%C3%A4isyyden-mukaan)

mikkoisa:
Looks good! Although nearest option should not be visible if user location is not available

PR #171, opened 2019-10-31
Improve unit contanct information

mikkoisa:
* Add text to unit contact information if a required field is missing
* Move Homepage link from e-services to contact information


mikkoisa:
[Toimipistetiedoille tieto aina näkyviin vaikka dataa ei olisi](https://trello.com/c/dfdhDMkN/260-toimipistetiedoille-tieto-aina-n%C3%A4kyviin-vaikka-dataa-ei-olisi)

src/views/UnitView/utils/unitSectionFilter.js
```diff
@@ -4,6 +4,7 @@ const unitSectionFilter = (list, section) => {
   let i = 0;
   list.forEach((item) => {
     if (!item.section_type) {
+      // TODO: is this used?
       filteredList.push({ type: section, value: item, id: i });
```
Dremars: Does this work properly? Seems like it adds items with `section_type` of null to list anyway. But this would mean a item without a `section_type` would be shown everywhere where `unitSectionFilter` is used.

mikkoisa: Must be some old implementation. Removed!

PR #175, opened 2019-11-07
Feature/new top bar

mikkoisa:
* New design for top bar (desktop & mobile)
* Modified document structure. New order for the start of the page is: h1(sr-only) -> jump link to main (sr-only) -> Top bar 
* Front page link will no longer reload the page. It pushes to home page and focuses to view title
* Modified the tabList top calculations, so that the sticky elements won't go under the top bar 

Note: Some elements have been commented out, since their functionality has not been implemented yet. The black tool bar looks slightly unfinished because of this.

### DIFF
--- a/src/components/TopBar/TopBar.js
+++ b/src/components/TopBar/TopBar.js
@@ -104,9 +104,9 @@ class TopBar extends React.Component {
                   <FormattedMessage id="general.frontPage" />
                 </Typography>
               </ButtonBase>
-              <Typography color="inherit">|</Typography>
+              <Typography aria-hidden color="inherit">|</Typography>
               {this.renderLanguages()}
-              <Typography color="inherit">|</Typography>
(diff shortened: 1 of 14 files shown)


src/components/TopBar/TopBar.js
```diff
@@ -1,184 +1,150 @@
...
+          <Toolbar className={classes.toolbarBlack}>
+            <div className={classes.toolbarBlackContainer}>
+              <ButtonBase role="link" onClick={() => this.navigateHome()}>
+                <Typography color="inherit">
+                  <FormattedMessage id="general.frontPage" />
+                </Typography>
+              </ButtonBase>
+              <Typography color="inherit">|</Typography>
+              {this.renderLanguages()}
+              <Typography color="inherit">|</Typography>
```
Dremars: These dividers should be aria-hidden or NVDA will enter empty element

PR #178, opened 2019-11-18
Feature/mobile side menu

mikkoisa:


src/components/TopBar/TopBar.js
```diff
@@ -115,20 +242,21 @@ class TopBar extends React.Component {
...
-              <HomeLogo mobile={type === 'mobile'} dark aria-hidden="true" className={classes.logo} />
+              <HomeLogo mobile={pageType === 'mobile'} dark aria-hidden="true" className={classes.logo} />
             </div>
             <MobileComponent>
               <div className={classes.mobileButtonContainer}>
                 {this.renderMapButton()}
-                {/* {this.renderMenuButton()} */}
+                {this.renderMenuButton()}
               </div>
+              {this.renderDrawerMenu(pageType)}
```
Dremars: Is it possible to set DrawerMenu as individual component?

mikkoisa: Good point! Will change it to an individual component, since it is not strictly related to TopBar anyway.

src/components/Navigator/Navigator.js
```diff
@@ -136,6 +136,24 @@ class Navigator extends React.Component {
...
+  openMap = () => {
+    const { history, location } = this.props;
+    const url = location.search ? `${location.search}&map=true` : '?map=true';
+    history.push(url);
+  }
+
+  // Remove map param from url
+  closeMap = (replace) => {
+    const { history, location } = this.props;
+    const url = location.pathname + location.search.replace('?map=true', '').replace('&map=true', '');
```
Dremars: Is there a reason for replace() being done twice?

Dremars:
Console is giving errror: `Failed prop type: Invalid prop 'settingsToggled' of type 'string' supplied to 'DefaultLayout', expected 'boolean'.`

Dremars:
Also menu button goes goes out of the window when choosing "All services" page from drawer menu.
MenuOutOfBound

src/components/DrawerMenu/DrawerMenu.js
```diff
@@ -0,0 +1,117 @@
...
+};
+
+DrawerMenu.propTypes = {
+  classes: PropTypes.objectOf(PropTypes.any).isRequired,
+  toggleSettings: PropTypes.func.isRequired,
+  currentPage: PropTypes.string.isRequired,
+  userLocation: PropTypes.objectOf(PropTypes.any).isRequired,
+  findUserLocation: PropTypes.func.isRequired,
+  intl: intlShape.isRequired,
+  settingsOpen: PropTypes.bool.isRequired,
```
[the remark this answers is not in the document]
mikkoisa: The prop error is temporary in this branch. When merged to develop it will go away.

src/components/Navigator/Navigator.js
mikkoisa: Yes, we need to replace both cases, since the param can be "?map=true" or "&map=true", depending if there is multiple params in the url or not. However this could be probably done with regex also.

Dremars: This seems to break search parameters in case where `?map=true` is replaced since it replaces `?` and thus rest of the parameters don't exist as search parameters anymore. For example /fi/search?map=true&q=kirjasto breaks when map is toggled on and off.

mikkoisa: Good point. Although in this case map=true is always the last parameter, but at some point it could cause problems. Will fix.

Dremars: Fixed in 9efb975

PR #182, opened 2019-12-03
Feature/new unit page style

mikkoisa:


src/components/SearchBar/styles.js
```diff
@@ -3,9 +3,10 @@ export default theme => ({
     color: theme.palette.primary.contrastText,
     display: 'flex',
     flexDirection: 'column',
-    backgroundColor: theme.palette.primary.main,
+    background: 'linear-gradient(340.58deg, #0B7BED 0%, #146CE8 67.04%, #1964E6 100%)',
```
Dremars: Is this linear-gradient same everywhere? If so we could put it in theme palette to `primary.gradient` variable.

src/components/TitleBar/TitleBar.js
```diff
@@ -51,18 +50,18 @@ const TitleBar = ({
 TitleBar.propTypes = {
   backButton: PropTypes.bool,
   classes: PropTypes.objectOf(PropTypes.any).isRequired,
-  title: PropTypes.string.isRequired,
+  title: PropTypes.oneOfType([PropTypes.string, PropTypes.objectOf(PropTypes.any)]).isRequired,
```
Dremars: Just noticed that `PropTypes.node` is supposed to represent anything that can be rendered. I think it would account for both of these instances. https://reactjs.org/docs/typechecking-with-proptypes.html

src/views/ServiceTreeView/ServiceTreeView.js
```diff
@@ -286,6 +286,7 @@ const ServiceTreeView = ({
           </List>
         </Collapse>
         <ServiceMapButton
+          text="Tee haku"
```
Dremars: Static text should be changed to use translations

mikkoisa: ServiceTreeView is not part of this PR (only the button prop here was changed). This is fixed on another branch

src/components/SearchBar/styles.js
mikkoisa: Yes! This gradient is used on many backgrounds. We should add this in another branch where we unify the new stylings.

src/components/TitleBar/TitleBar.js
mikkoisa: Really good point! I think this PropTypes.node could be used in several other components as well.

PR #201, opened 2020-01-07
Fix/unit view data

mikkoisa:
- Fix issues caused by previous modifications to unit view
- Move unit event list redux state to selectedUnit redux state (where reservations are already)
- Further improve unit view lifecycle methods
- Move selected unit state modification to unit view from other components


src/components/ListItems/SimpleListItem/SimpleListItem.js
```diff
@@ -41,11 +41,9 @@ const SimpleListItem = (props) => {
           </Typography>
         </ListItemText>
       </ListItem>
-      {divider ? (
-        <li>
```
Dremars: I'm wondering if having `<li>` element itself as aria hidden would solve problem with list length being wrong with screenreaders.

src/redux/actions/selectedUnitEvents.js
```diff
@@ -0,0 +1,30 @@
...
+  const onStart = () => dispatch(more ? isFetchingMore() : isFetching());
+  const onSuccess = data => dispatch(fetchSuccess(data.data, data.meta));
+  const onError = e => dispatch(fetchError(e.message));
+
+  // Fetch data
+  unitEventsFetch({ location: `tprek:${unitId}`, page_size: pageSize || 5 }, onStart, onSuccess, onError);
+};
+
+
+export const fetchAdditionalEvents = (unitId, pageSize, next) => async (dispatch) => {
```
Dremars: UnitId and pageSize seem unnecessary since next parameter is passed to `fetchWrapper` and overrides all options with `null`. In this case ```{ location: `tprek:${unitId}`, page_size: pageSize }``` is unnecessary.

src/redux/actions/selectedUnitReservations.js
```diff
@@ -2,21 +2,34 @@ import { reservations } from './fetchDataActions';
...
   };
   const onError = e => dispatch(fetchError(e.message));
 
   // Fetch data
-  reservationsFetch({ unit: `tprek:${id}` }, onStart, onSuccess, onError, null, id);
+  reservationsFetch({ unit: `tprek:${id}`, page_size: pageSize || 5 }, onStart, onSuccess, onError);
+};
+
+
+export const fetchAdditionalReservations = (id, pageSize, next) => async (dispatch) => {
```
Dremars: Id and pageSize seem unnecessary since next parameter is passed to `fetchWrapper` and overrides all options with `null`. In this case ```{ unit: `tprek:${unitId}`, page_size: pageSize }``` is unnecessary.

src/redux/reducers/fetchDataReducer.js
```diff
@@ -61,8 +61,10 @@ const dataSetReducer = (state, action, prefix) => {
 // Reducers for fetching single data
 const dataSingleInitialState = {
   isFetching: false,
+  isFetchingMore: false,
```
Dremars: I think we can have single `isFetching` value in state. We either are fetching or not. Otherwise you have to check both values everywhere since we can't start new fetches if data is currently being fetched in any way.

src/redux/reducers/fetchDataReducer.js
```diff
@@ -73,6 +75,12 @@ const dataSingle = (state, action, prefix) => {
         isFetching: true,
         errorMessage: null,
       };
+    case `${prefix}_IS_FETCHING_MORE`:
```
Dremars: This can be removed if we move back to single `ìsFetching` lets handle differences in redux actions.

src/components/ListItems/SimpleListItem/SimpleListItem.js
mikkoisa: Works good that way! Fixed

src/redux/actions/selectedUnitEvents.js
mikkoisa: Good point, fixed!

src/redux/actions/selectedUnitReservations.js
mikkoisa: Fixed

src/redux/reducers/fetchDataReducer.js
mikkoisa: This is true, fixed!

src/redux/reducers/fetchDataReducer.js
mikkoisa: Fixed

PR #202, opened 2020-01-07
Fix/search suggestion zindex

Dremars:
- Fix issue with z-index in address view search suggestions
- Refactor z-indexes to theme object


Dremars:
[Yläosion sticky elementit rikkoutuvat](https://trello.com/c/ebgSJR3u/269-yl%C3%A4osion-sticky-elementit-rikkoutuvat)

src/layouts/DefaultLayout.js
```diff
@@ -44,6 +44,10 @@ const createContentStyles = (
       visibility: isMobile && (!mobileMapOnly || settingsOpen) ? 'hidden' : '',
       height: isMobile ? `calc(100% - ${topBarHeight})` : '100%',
       width: '100%',
+      zIndex: 1300,
```
mikkoisa: Näyttökuva 2020-1-8 kello 15 03 11
This z-index of the map seems to be larger than top bar's

src/views/MapView/styles.js
```diff
@@ -127,10 +127,10 @@ const styles = theme => ({
     color: '#fff',
   },
   topArea: {
-    zIndex: 99999999,
     position: 'fixed',
     top: config.topBarHeight,
     width: '100%',
+    zIndex: theme.zIndex.modal,
```
mikkoisa: Could this z-index use theme.zIndex.infront for example, since the element is not actually modal

src/layouts/DefaultLayout.js
Dremars: Good point. This should be fixed now

src/views/MapView/styles.js
Dremars: True. I've changed it to infront in 2d92484

PR #217, opened 2020-01-15
Fix/list item stylings

Dremars:
- Fix ResultItem and SimpleListItem stylings
- Add style variation option for ResultItem

### DIFF
--- a/src/components/ListItems/ResultItem/ResultItem.js
+++ b/src/components/ListItems/ResultItem/ResultItem.js
@@ -22,7 +22,7 @@ const ResultItem = ({
   srLabel,
   intl,
   selected,
-  single,
+  padded,
   extendedClasses,
 }) => {
   // Screen reader text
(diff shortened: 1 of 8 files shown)


Dremars:
[Tapahtumasivun korjaukset](https://trello.com/c/1aeLHfIQ/322-tapahtumasivun-korjaukset)

mikkoisa:
Also the left side padding in services in search results page should be checked.
Näyttökuva 2020-1-16 kello 13 03 53

mikkoisa:
These list items are also a bit too much to the left (they should be on the right side of the title)
Näyttökuva 2020-1-16 kello 13 08 31

Dremars:
@mikkoisa I've unified simple list items and result items to have unified styling and aligned items with TitledList title. List item hover area doesn't fill whole width because parent container has padding. Fixing this would require checking whole page layout. This is also issue in current version and should be checked separately.

PR #218, opened 2020-01-16
Add high contrast map option

mikkoisa:
Add new high contrast map and update default map styles.


mikkoisa:
[Kontrastikartan lisäys](https://trello.com/c/aubRRlk0/344-kontrastikartan-lis%C3%A4ys)

mikkoisa:
Map now changes to high contrast map only when changing theme from default to dark.

PR #232, opened 2020-01-22
Feature/marker clusters

Dremars:
New markercluster feature
- Changed marker drawing to use leaflet rather than react-leaflet


Dremars:
[Marker clusterit käyttöön](https://trello.com/c/kowYpMje/363-marker-clusterit-k%C3%A4ytt%C3%B6%C3%B6n)

Dremars:
> Works really well and feels smooth to use! For further reference: some markers that overlap cluster markers could probably be merged to cluster. But we should go with this version for now.
> Näyttökuva 2020-1-22 kello 14 03 21

Added new card to To Do list for this.

PR #244, opened 2020-01-25
Add wrapper div to fix layout issues with safari

mikkoisa:
* NOTE: also temporarily disabled min height for tabs, since it caused some issues.

### DIFF
--- a/src/components/TabLists/TabLists.js
+++ b/src/components/TabLists/TabLists.js
@@ -240,7 +240,7 @@ class TabLists extends React.Component {
       && typeof tabsHeight === 'number'
     ) {
       // Adjust tabs min height to work with shorter contents
-      const customTabHeight = containerHeight - tabsHeight;
+      // const customTabHeight = containerHeight - tabsHeight;
       // Set new styles and scrollDistance value to state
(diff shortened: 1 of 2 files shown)


mikkoisa:
[Toimipistesivun sticky elementit rikkoutuvat pitkällä välilehdellä (vain safarilla?)](https://trello.com/c/ZJCLBsyw/365-toimipistesivun-sticky-elementit-rikkoutuvat-pitk%C3%A4ll%C3%A4-v%C3%A4lilehdell%C3%A4-vain-safarilla)

mikkoisa:
Based on the comments, it seems that the idea behind the min height was that the scroll on the page wouldn't jump too much when changing from a long tab to a really short one. (Issue is mostly on mobile). I too prefer removing it.

Dremars:
> Based on the comments, it seems that the idea behind the min height was that the scroll on the page wouldn't jump too much when changing from a long tab to a really short one. (Issue is mostly on mobile). I too prefer removing it.

I can't replicate this even on mobile. Seems like it's already fixed anyway. When changing tabs tab returns to top either way.

PR #256, opened 2020-01-27
Feature/info view

mikkoisa:


src/i18n/translations/sv.js
```diff
@@ -107,6 +108,7 @@ export default {
   'general.pageTitles.address': 'Adressvy',
   'general.pageTitles.list.events': 'Förteckning över evenemang ',
   'general.pageTitles.list.reservations': 'Förteckning över reserveringar ',
+  'general.pageTitles.info': 'Infovy',
```
Dremars: This doesn't seem correct

PR #257, opened 2020-01-27
Change address data fetching to fetch data from address endpoint

Dremars:
Change address url data to correct data on when address language doesn't represent current language


Dremars:
[Kieli-parametri osoitetietojen hakuun](https://trello.com/c/5W9iHJIu/388-kieli-parametri-osoitetietojen-hakuun)

PR #269, opened 2020-01-30
Fix search page settings link accessibility

mikkoisa:
* Use button component so that key presses are handled properly
* Remove h3 from settings link, since there is already a h3 before it


mikkoisa:
[Hakutulossivun saavutettavuus](https://trello.com/c/zj917qtC/374-hakutulossivun-saavutettavuus)

PR #277, opened 2020-02-03
Fix/banner accessibility

mikkoisa:
* Settings page is now dialog that traps keyboard and screen reader users inside, so that it's only possible to leave the dialog by pressing one of the close buttons. Keyboard navigation loops through the dialog, sr navigation makes rest of the page aria-hidden.

### DIFF
--- a/src/components/Settings/Settings.js
+++ b/src/components/Settings/Settings.js
@@ -35,6 +35,7 @@ class Settings extends React.Component {
 
   constructor(props) {
     super(props);
+    this.dialogRef = React.createRef();
     this.state = {
       alert: false,
       currentSettings: {},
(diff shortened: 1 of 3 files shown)


mikkoisa:
I found that this implementation for settings dialog is most logical and robust. Other solutions could be discussed as well.

mikkoisa:
[Yläpalkki saavutettavuus](https://trello.com/c/iLHvuwd2/397-yl%C3%A4palkki-saavutettavuus)

src/components/Settings/Settings.js
```diff
@@ -614,12 +627,14 @@ class Settings extends React.Component {
...
           <SMButton
             small
+            role="button"
             messageID="general.save"
             onClick={() => this.saveSettings()}
             color="primary"
           />
           <SMButton
             small
+            role="button"
```
Dremars: I think SMButton should have button role by default and then change depending on props. But lets change this independently from this PR.

mikkoisa: Yes, I also found this surprising that the button needs to have role button added. Also screen reader looping in the dialog was not implemented, since it would be confusing to the user (blind users can't know that it loops, instead it feels that the dialog is infinitely long!)

PR #282, opened 2020-02-04
Front page accessibility

mikkoisa:
* Add new disabled style to home buttons
* Fix missing hover styles
* Modify home page icons to Inline SVGs so that the icon color can be changed easily
* Make home page icons smaller


mikkoisa:
[Etusivu saavutettavuus](https://trello.com/c/dFO7JxjO/407-etusivu-saavutettavuus)

src/components/PaperButton/styles.js
```diff
@@ -4,6 +4,25 @@ export default theme => ({
     padding: 0,
     width: '88%',
     maxWidth: 356,
+    '&:hover': {
+      backgroundColor: theme.palette.primary.main,
+      '& $text': {
+        color: '#fff',
```
Dremars: I think we should use relative text colors. For example if `primary.main` becomes lighter color white might not have enough contrast. We should try using `primary.contrastText` variable or if that's not good lets add another variable `primary.starkContrast` or something.

src/components/PaperButton/styles.js
```diff
@@ -4,6 +4,25 @@ export default theme => ({
...
+      },
+      '& $iconContainer': {
+        backgroundColor: '#fff',
```
Dremars: Same as above lets use relative colors using theme.

src/components/PaperButton/styles.js
```diff
@@ -4,6 +4,25 @@ export default theme => ({
...
+        color: theme.palette.primary.main,
+        '& g': {
+          fill: theme.palette.primary.main,
+        },
+      },
+    },
+  },
+  disabled: {
+    pointerEvents: 'none',
+    backgroundColor: '#efefef',
```
Dremars: Does this background color take into account the subtitle text. Not sure if the contrast is good enough. Wasn't able to test this though since texts are using rgba values with reduced alpha. I think we could move to using hexcolors so contrast can be checked more easily.

Also contrast with icon button container seems low. Maybe we can check this together with Nina

mikkoisa: Good call. I checked the contrast ratio value with tool that takes alpha value into account and the ratio is too low to pass wcag AAA standards. Will change to darker text.

src/components/PaperButton/styles.js
mikkoisa: Fixed!

src/components/PaperButton/styles.js
mikkoisa: Fixed!

PR #287, opened 2020-02-06
Feature/bbox parameter option

Dremars:
Handle bbox parameters for unit fetching and map focusing. Add independent DataFetcher component for easier fetch handling.

Meant to solve problems with old embeds and links. For example 
- `/embed/unit?service=25656&bbox=60.10423,24.57757,60.24830,25.12205&city=helsinki`
- `/embed?level=all&bbox=60.184077,24.978499,60.186077,24.980499`
- `/embed?lang=fi&bbox=60.25251,25.21889,60.28112,25.32483&level=all`

Should also handle fetching etc without embed in url.

Should test properly that everything works. There was some problems with old bbox functionality for example old bbox used another coordinate projection to get data from larger area. But this data didn't fall under given bbox parameter so I think it's still fine to exclude that extra data in this new solution.

Also old version had issues where bbox focusing was overridden by focus to unit functionality. This means that in new version `embed/unit?service=25656&bbox=60.10423,24.57757,60.24830,25.12205&city=helsinki` shows the actual bbox area given rather than trying to focus by fitting units to screen.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -19,6 +19,7 @@ import { changeLocaleAction } from './redux/actions/user';
 import DefaultLayout from './layouts';
 import EmbedLayout from './layouts/EmbedLayout';
 import Navigator from './components/Navigator';
+import DataFetcher from './components/DataFetchers/DataFetcher';
 
 class App extends React.Component {
   constructor(props) {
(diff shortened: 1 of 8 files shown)


Dremars:
[Upotuksen bbox rajaus ja level ei toimi](https://trello.com/c/n0ECZmVq/394-upotuksen-bbox-rajaus-ja-level-ei-toimi)

mikkoisa:
Seems to be working well, didn't find any issues while testing!

PR #296, opened 2020-02-12
Sort service tree nodes by ID

mikkoisa:


tituomin:
Unfortunately the problem this is trying to fix cannot be fixed client-side. The request was that the ordering should respect the order of the json array element returned by the TPR API. That is not the same as the numerical sort order of the ids. The array indices need to be persisted in the servicemap backend and then the data returned in the same order from the api. No UI changes necessary, I think.

PR #308, opened 2020-02-14
Service tree accessibility

mikkoisa:
* Accessibility: added indeterminate value to checkboxes
* Fix issue where node wouldn't be selected if all child nodes are selected


mikkoisa:
Still needs to be checked that NVDA reds indeterminate checkboxes correctly

PR #313, opened 2020-02-18
Release v1.0.14

mikkoisa:
Currently in pre-release
- Opening settings page from search view settings link will now put focus correctly on top of settings dialog. Closing settings will put focus back to the link.
- Settings page is now visually on top of sidebar instead of replacing sidebar content
- Improve Tab-element focus indicator
- Service tree: added indeterminate value to checkboxes
- Service tree: Fix issue where node wouldn't be selected if all child nodes are selected

### DIFF
--- a/src/layouts/DefaultLayout.js
+++ b/src/layouts/DefaultLayout.js
@@ -56,9 +56,14 @@ const createContentStyles = (
       bottom: 0,
       width,
       margin: 0,
-      overflow: !isMobile ? 'auto' : '',
+      // eslint-disable-next-line no-nested-ternary
+      overflow: settingsOpen ? 'hidden'
+        : isMobile ? '' : 'auto',
       visibility: mobileMapOnly && !settingsOpen ? 'hidden' : null,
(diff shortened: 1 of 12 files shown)


src/layouts/DefaultLayout.js
```diff
@@ -56,10 +56,15 @@ const createContentStyles = (
       bottom: 0,
       width,
       margin: 0,
-      overflow: !isMobile ? 'auto' : '',
+      // eslint-disable-next-line no-nested-ternary
+      overflow: settingsOpen ? 'hidden'
+        : isMobile ? '' : 'auto',
```
Dremars: Does this mean that mobile checks are only done if settings isn't open? Does this have a purpose?

mikkoisa: Yes, this operaation looks really unlogical (probably why eslint warns about nesting ternaries). The idea is, that If settings are open, both layouts have overflow hidden. If settings are not open, use the normal values (which are ' ' for mobile and 'auto' for desktop)

PR #315, opened 2020-02-18
Prevent minimization on server side

Dremars:


Dremars:
Need to check that nothing breaks

tituomin:
> If you add one more line ...
> 
> ```
>       minimize: false,
>       concatenateModules: false
>   },
> ```
> 
> Then we can get rid of the "Error while fetching data: Expected signal to be an instanceof AbortSignal" errors which are currently filling the production logs and make debugging a bit harder.

Explanation: webpack, when concatenating js modules, renames the AbortSignal class name to abort_controller_AbortSignal, which breaks the (admittedly stupid) check node-fetch is making.

Dremars:
> > If you add one more line ...
> > ```
> >       minimize: false,
> >       concatenateModules: false
> >   },
> > ```
> > 
> > 
> > Then we can get rid of the "Error while fetching data: Expected signal to be an instanceof AbortSignal" errors which are currently filling the production logs and make debugging a bit harder.
> 
> Explanation: webpack, when concatenating js modules, renames the AbortSignal class name to abort_controller_AbortSignal, which breaks the (admittedly stupid) check node-fetch is making.

Yeah and I think this can be changed back once node-fetch gets new release and we can update it. But now everything seems to work fine.

PR #316, opened 2020-02-19
Fix focus loss on settings confirmation click

mikkoisa:
Previously clicking save or cancel on settings page notification box would make screen readers stuck. (Focus is nowhere). Now it focuses back to settings title.


mikkoisa:
[Tallennusilmoitus saavutettavuus](https://trello.com/c/bsy67jkG/447-tallennusilmoitus-saavutettavuus)

PR #317, opened 2020-02-19
Add correct focus contrast to top area buttons

mikkoisa:
- Service tree blue area buttons used default focus (blue) instead of correct one (white)


mikkoisa:
[Palvelupuu yläosan näppäimistöfokus](https://trello.com/c/qASZE4MG/448-palvelupuu-yl%C3%A4osan-n%C3%A4pp%C3%A4imist%C3%B6fokus)

PR #324, opened 2020-02-20
Fix/embed search redirect

Dremars:
- Maintain embed on search redirect
- Embed overlay

### DIFF
--- a/src/i18n/translations/fi.js
+++ b/src/i18n/translations/fi.js
@@ -46,7 +46,7 @@ export default {
   'event.title': 'Tapahtumat',
 
   // Embed
-  'embed.click_prompt_move': 'Klikkaa siirtyäksesi palvelukartalle',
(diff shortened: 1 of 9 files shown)


src/i18n/translations/fi.js
```diff
@@ -45,6 +45,9 @@ export default {
   'event.picture': 'Tapahtuman kuva',
   'event.title': 'Tapahtumat',
 
+  // Embed
+  'embed.click_prompt_move': 'Klikkaa siirtyäksesi palvelukartalle',
```
tituomin: Maybe "Palvelukartalle" capitalized.

Dremars:
> Functionality works well and the logo link is nice. Clicking a cluster on embed view opens Servicemap on new tab and opens the cluster in the embed view. This should be fixed at some point if unintentional, however it could be part of different PR as well.

Should be fixed now with 55036b6

PR #330, opened 2020-02-21
Add tabpanel role and tab controls

Dremars:
Add tabpanel role and controls to match WCAG standards.


Dremars:
[Tab-panel rooli välilehdille](https://trello.com/c/3XHr1tME/422-tab-panel-rooli-v%C3%A4lilehdille)

PR #333, opened 2020-02-24
Feedback form accessibility

mikkoisa:
* Add better outlines for focused textfields.
* Add outline styles for contrast view.
* Fix element positions and styles
* Add "sending" state to feedback page. The state will disable send button while feedback is being sent.
* Add better screen reader information about missing required fields: sr-only alert message and sr-label to disabled button
* Fix link functionality (now space opens link as well) and style (link now has underline so it looks like link in contrast theme as well)
* Improve button contrasts


mikkoisa:
[Palautelomake saavutettavuus](https://trello.com/c/yEDhiALl/450-palautelomake-saavutettavuus)

PR #335, opened 2020-02-27
Feature/unit-events-and-reservations-page

Dremars:
Added own pages for unit reservations and events for better accessibility. Old event listing didn't work well for screen readers because lists were too long and made the page way too long with lists. New pages show data with paginated lists which should be better for screenreaders

- New list item component: EventItem 
- New list item component: ReservationItem
- New PaginatedList component
- ExtendedData component (handles viewing events and reservations)
- Changes to unit events and reservation fetching
- Unit events and reservation state structure changed to dataSet


Dremars:
Need to make sure that it's clear to users that new pages are subpages for unit view. Also check that pages and transitions are clear and accessible.

Dremars:
[Toimipistesivu: tapahtumille omat sivut](https://trello.com/c/3XWRJqmv/380-toimipistesivu-tapahtumille-omat-sivut)

mikkoisa:
Reservation count breaks when returning from new reservations page back to unit page. However seems like it does not happen on some units (maybe units with large amount of reservations?)

mikkoisa:
When reloading the unit page, events count breaks as well.
Näyttökuva 2020-2-27 kello 16 45 25

Dremars:
> When reloading the unit page, events count breaks as well.
> Näyttökuva 2020-2-27 kello 16 45 25

Server side fetches didn't set count and max values. Fixed in 00f08a5

Dremars:
> Reservation count breaks when returning from new reservations page back to unit page. However seems like it does not happen on some units (maybe units with large amount of reservations?)
> Näyttökuva 2020-2-27 kello 16 45 11
> 
> Example that breaks: /fi/unit/15311?p=1&t=2

This seems to happen in cases where data is fetched with `onNext` option but `page_size` is set to larger than total number of results. This was onNext function was ignored. I've changed onNext to run before returning results. be3b35b

PR #336, opened 2020-02-27
Mobile accessibility

mikkoisa:
- Added correct aria-labeling to drawer menu button
- Modifed button components so that the text inside the button is aria-hidden, and the button itself has the text as aria-label (iOS mobile VoiceOver would only read the text inside the button ignoring the button and its state completely)
- Fixed issue with  focus disappearing from page on entering settings page
- Added aria-hidden to "show on map" -button on unit page
- Added "search" -type to search field, so that virtual keyboard has correct layout


mikkoisa:
[Saavutettavuus mobiililla](https://trello.com/c/RCPSKVij/377-saavutettavuus-mobiililla)

PR #337, opened 2020-02-27
Add unit cluster hover elements based on zoom depth

Dremars:
New unit cluster hover element. Content is based on zoom level:
- When zoomed close cluster hover shows sticky popup with list of units within cluster.
- When zoomed out cluster hover shows a tooltip like popup with simple info of how many units exists within cluster.


Dremars:
[Marjaryppäiden hover popup](https://trello.com/c/TN9oAASe/443-marjarypp%C3%A4iden-hover-popup)

mikkoisa:
The distance of markers in address view should probably use the location of the address.

Näyttökuva 2020-3-6 kello 10 12 01

PR #338, opened 2020-02-27
Improve expanded suggestions page accessibility

mikkoisa:
- Fixed broken screen reader focus when entering expanded suggestions page
- Fixed incorrect screen reader notification about amount of suggestions
- Fixed incorrect text on title bar back button (previously said "back to home page" instead of "back to search"
 - Added backButtonText prop to title bar back button to override default back button text
- Moved "Refine search"- button from Expanded suggestions view to Search view, so that it can be focused when returning to Search view


mikkoisa:
[Tarkennettu haku -näkymän saavutettavuus](https://trello.com/c/8ggxQ6JY/462-tarkennettu-haku-n%C3%A4kym%C3%A4n-saavutettavuus)

PR #342, opened 2020-03-04
Feature/coordinate parameter

Dremars:
New feature to allow custom location coordinate for service view. Coordinate will be used to render user marker to custom position and using this location to calculate unit distances. Units will automatically be listed based on distance if coordinates are given. 
Example `/fi/service/813?latLng=60.281936%2C24.949933`


Dremars:
Changed parameters to use `lat=60.281936&lon=24.949933` values and option to hide user marker with `hide_marker=true`

PR #350, opened 2020-03-11
Feature/update service view

Dremars:


mikkoisa:
I think that the service icons in serviceItem component should be updated to the new one as well (icons on service search results and unit page service icons). This would require a new dark version of the icon .

Dremars:
Yeah I asked about this from Nina she thought that list items could stay the same. But these icons seem fine as well and more consistent

Dremars:
[Palvelusivu rakenne](https://trello.com/c/a9cGlVHW/376-palvelusivu-rakenne)

PR #353, opened 2020-03-16
Optimize initial load

mikkoisa:
- Split Webpack bundle into smaller chunks using [dynamic import](https://reactjs.org/docs/code-splitting.html#import) with [loadable-components](https://loadable-components.com)

- Compress bundle to gzip 

Currently all of the views have been split from main bundle into separate chunk named views.js. Likewise settings component and search bar component have been split from main to components.js. Main bundle size can be further reduced by splitting additional components into separate chunks.


mikkoisa:
Subsequent loads still sometimes crash markerCluster component. Could be that the load from cache is too fast so that the markerCluster tries to add layers to map before it has loaded correctly?

webpack.config.js
```diff
@@ -40,7 +41,7 @@ const js = {
 };
 
 const fonts = (isClient = true) => ({
```
Dremars: Maybe we should change this to assets because this is using file-loader to get images as well

server/server.js
```diff
@@ -25,6 +26,9 @@ import legacyRedirector from './legacyRedirector';
 import matomoTrackingCode from './analytics';
 import { getLastCommit, getVersion } from './version';
 
+// This is the stats file generated by webpack loadable plugin
+const statsFile = path.resolve('./dist/src/loadable-stats.json')
```
Dremars: Does this resolve correctly? Seems to get /dist from current folder.

mikkoisa: Yes it seems to resolve correctly. Originally I did use ../dist but that caused errors.

PR #354, opened 2020-03-17
Release v1.0.22

mikkoisa:
- Refactored distance calculations
- Update service view stylings
- New cluster hover popups and added extra info to marker tooltips
- Update transit stop info styles


Dremars:
If user doesn't have location on and goes from address page back to search. Map markers are showing distance based on address rather than none. Distance calculation shouldn't use address if not on address page

PR #356, opened 2020-03-17
Fix/feedback link

mikkoisa:
- Display unit feedback button only for units with contract types municipal_service or purchased_service
- Update device wrapper component to allow null children as props


mikkoisa:
[Palutelinkki pitää katsoa oraganisaatiosta:](https://trello.com/c/qeV7Nk5P/474-palutelinkki-pit%C3%A4%C3%A4-katsoa-oraganisaatiosta)

PR #358, opened 2020-03-18
LocalStorageUtility add try-catch to when setting localStorage

Dremars:
Prevents crash when user has blocked cookies on browser.


Dremars:
[Uncaught DOMException: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.](https://trello.com/c/xDXmZOmv/485-uncaught-domexception-failed-to-read-the-localstorage-property-from-window-access-is-denied-for-this-document)

PR #361, opened 2020-03-23
New regression tests for search view

Dremars:
Add navigation tests for search suggestions
Add regression test for search view


Dremars:
Missing tests for search suggestion arrow key navigation.
- Can search suggestions be navigated with arrow keys
- Does search suggestion navigation loop correctyl

PR #368, opened 2020-04-20
Feature/address view area update

Dremars:
Close-by services tab for address view.
Link component for unified text link stylings


Dremars:
Waiting for missing translations.

Dremars:
Need to change which markers are drawn on map based on selected tab

Dremars:
Map drawing updated in 7bf7065

src/components/ListItems/DivisionItem/DivisionItem.js
```diff
@@ -0,0 +1,154 @@
...
+    : ''} 
+  `;
+
+  return (
+    <>
+      <ListItem
+        component="li"
+        className={classes.listItem}
+      >
+        <SMLink
```
mikkoisa, line 53: Could SMLink-component be replaced by Mui ButtonBase-component? ButtonBase with styles: flexDirection: 'column' and alignItems: 'initial' seems to be working identically.

Dremars: Changed in fecf6c7

Dremars:
> The font size and style look a little different compared to the design layout. Also looks like the names should be of type "Neuvola" instead of "Neuvola-alue" .
> Näyttökuva 2020-4-29 kello 13 20 07

We talked about names earlier in meetings and decided it's fine with area text. Style can't be exactly same since original design didn't give direction for links.

mikkoisa:
Ok, looks good!

PR #369, opened 2020-04-21
Add hearing maps to unit view

mikkoisa:


mikkoisa:
Currently only Malmitalo has hearing maps available.

src/views/UnitView/UnitView.js
```diff
@@ -180,6 +185,21 @@ const UnitView = (props) => {
 
     return (
       <div className={classes.content}>
+        {hearingMaps && (
+          <TitledList title="Kuulokartat">
```
Dremars, line 189: TitledList places h3 element by default. But sibling headings are using h4. Use `titleComponent="h4"`

mikkoisa: Fixed

PR #371, opened 2020-04-23
Correct sticky styles to title bar

mikkoisa:
**Make TitleBar component sticky on most of the pages**
- Update TitleBar component so that the component itself contains the sticky styles.
- Add new hook _useMobileStatus_  which makes mobile layout check easier
- Update TabList component sticky calculations


Dremars:
There is some issues with long contents for example `fi/event/helmet:210051` sticky works at start but disappears when scrolled to bottom.

Also noticed that event- and reservations listing does have TitleBar elements which I think should be sticky as well. `/fi/unit/51342/reservations` and `/fi/unit/51342/events`

mikkoisa:
The sticky element disappearing on long pages can be fixed by adding div around the content, will fix it on the event page!

I will add sticky to event- and reservations list pages as well.

PR #393, opened 2020-05-26
Feature/area view

mikkoisa:
* Note: translations still missing


src/views/AreaView/components/UnitTab/UnitTab.js
```diff
@@ -0,0 +1,161 @@
...
+
+      return (
+        <div>
+          <div className={`${classes.subtitle} ${classes.sidePadding}`}>
+            <AddressIcon className={`${classes.rightPadding} ${classes.addressIcon}`} />
+            <Typography className={classes.selectedAddress}>
+              {formAddressString(selectedAddress)}
+            </Typography>
+          </div>
+          <Divider />
```
Dremars, line 103: Add aria-hidden to divider

src/views/AreaView/components/AreaTab/AreaTab.js
```diff
@@ -0,0 +1,188 @@
...
+
+
+  const renderCollapseContent = (item) => {
+    if (fetching.includes(item.title)) {
+      return (
+        <Typography aria-live="polite">
+          <FormattedMessage id="general.loading" />
+        </Typography>
+      );
+    }
```
Dremars, line 63: Since this element disappears the order for screen reader changes. If user is on loading element and new list of inputs is inserted screen reader thinks user is at 1st item of that list and next element will be 2nd. 
We could keep this element always existing and change content accordingly

Also for NVDA aria-live functionality is not triggered for some reason when this element appears.

Dremars: Aria-live problem could be caused by this whole element appearing rather than the inner content of aria-live

src/views/AreaView/components/UnitTab/UnitTab.js
mikkoisa: Fixed

src/views/AreaView/components/AreaTab/AreaTab.js
mikkoisa: Added separate screen reader alert message to the bottom of the page. This alerts "loading" when any fetch is started, and "loading finished" when all fetches are finished. This however leaves the alert message to the bottom of the page that can be navigated to.

mikkoisa: Since the message will be "loading finished", I don't think that it is too harmful if the user navigates to it. Some trick with setTimeout can be done, so that the alert box disappears shortly after reading the message, but i'm not sure if it is necessary.

PR #400, opened 2020-06-03
Feature/button unit tests

Dremars:
Change ServiceMapButton srText to aria-label to keep these values consistent and avoid confusion with prop naming. 

Add unit tests for ServiceMapButton component


Dremars:
[Button komponentin yksikkötestit](https://trello.com/c/Y1MnAH5T/518-button-komponentin-yksikk%C3%B6testit)

PR #401, opened 2020-06-05
Fix TabList font sizes on small screens

mikkoisa:
Problem occurred on two cases: on small tablet screen and on small mobile screen (with Swedish language selected)
Fixed by modifying font size to be relative to view width, with min and max values.


mikkoisa:
[Tomipistesivun välilehtien otsikoihin väliä](https://trello.com/c/vezs3Zmn/531-tomipistesivun-v%C3%A4lilehtien-otsikoihin-v%C3%A4li%C3%A4)

Dremars:
Seems like the problem still occurs when scaling to smaller screens.
tab-problem

 I  wonder if tab elements should have some horizontal padding and maybe break words to avoid texts overflowing. But this means that text will break by cutting couple letters to next line.

PR #419, opened 2020-06-26
Change PaperButton to adjust to container width

Dremars:
Fix issue with home view paper buttons not staying in line with searchbar.


Dremars:
[Tasataan etusivun elementit](https://trello.com/c/ryV9iegX/568-tasataan-etusivun-elementit)

PR #424, opened 2020-07-03
Prevent map clustering from changing when list ordering is changed

Dremars:
Prevent map clustering changes when unit ordering is changed by user. Previously map markers were re-rendered every time user changed the order priority for unit listing. This caused list data to change and affected the map clustering by changing cluster positions.

This solution introduces new getProcessedMapData selector which completely ignores ordering the data.

### DIFF
--- a/src/redux/selectors/results.js
+++ b/src/redux/selectors/results.js
@@ -11,13 +11,14 @@ const cities = state => [
   ...state.settings.kauniainen ? ['kauniainen'] : [],
 ];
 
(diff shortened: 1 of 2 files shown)


Dremars:
[Kartan marjaklusterit muuttuvat kun hakutulosten järjestystä muuttaa](https://trello.com/c/S31NClP0/565-kartan-marjaklusterit-muuttuvat-kun-hakutulosten-j%C3%A4rjestyst%C3%A4-muuttaa)

PR #461, opened 2020-09-03
Download tool + memoized selector refactor

mikkoisa:
- Refactor reselect selectors to work correctly by removing state passing: previously state was given as parameter to selector, this causes memoization to not work.
- Update react-redux to get new redux hook useSelector, that is used in download tool hook.
- Add download tool hook
- Changed tool menu styles

**Note**: new redux hooks allows easier access to state on functional components. These can be used to replace the connect HOC and mapStateToProps.

### DIFF
--- a/src/utils/orderUnits.js
+++ b/src/utils/orderUnits.js
@@ -1,0 +1,90 @@
+import UnitHelper from './unitHelper';
+import { getLocaleString } from '../redux/selectors/locale';
+import { calculateDistance } from '../redux/selectors/unit';
+
+const orderUnits = (unitData, sortingParameters) => {
+  const {
+    usedPosition, direction, order, locale, settings,
+  } = sortingParameters;
+
+  let results = Array.from(unitData);
+
+  switch (order) {
+    // Accessibility
+    case 'accessibility': {
+      results.forEach((element) => {
+        // eslint-disable-next-line no-param-reassign
+        element.shorcomingCount = UnitHelper.getShortcomingCount(element, settings);
+      });
+      results.sort((a, b) => {
+        if (a.object_type !== 'unit' || b.object_type !== 'unit') {
+          return 0;
+        }
+        const aSC = a.shorcomingCount;
+        const bSC = b.shorcomingCount;
+
+        if (aSC === null || (aSC === null && bSC === null)) { return -1; }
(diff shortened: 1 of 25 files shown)


mikkoisa:
Still todo: add download tool to tool menu when it's available

src/views/AddressView/utils/fetchAddressUnits.js
```diff
@@ -6,6 +6,8 @@ const fetchAddressUnits = async (lnglat) => {
     lon: `${lnglat[0]}`,
     distance: 500,
     only: 'name,location,accessibility_shortcoming_count,',
+    include: '',
```
Dremars, line 9: What's this meant to do?

src/utils/orderUnits.js
```diff
@@ -0,0 +1,90 @@
...
+
+  let results = Array.from(unitData);
+
+  switch (order) {
+    // Accessibility
+    case 'accessibility': {
+      results.forEach((element) => {
+        // eslint-disable-next-line no-param-reassign
+        element.shorcomingCount = UnitHelper.getShortcomingCount(element, settings);
+      });
```
Dremars, line 18: I'm wondering if this could be moved near fetching so it's not run each time order is change but rather when data is first time fetched. 

In the future I think we might need to have some kind of handler to process each data type (unit, address, etc.) before inserting it to redux.

Dremars (commented):
This works way better than the previous implementation. Good job!

src/views/AddressView/utils/fetchAddressUnits.js
mikkoisa: Used to override default include value, this was not supposed to be included in the commit. Will remove.

src/utils/orderUnits.js
mikkoisa: Good point. However shortcoming counting requires user settings, which can change after the fetch, so this needs to be recalculated each time user settings or unit data changes.

Dremars: Good point

Dremars (changes requested):
Seems like marker rendering fails when selecting unit from search and returning back to search page. Units are not updated.

Dremars (approved):

PR #429, opened 2020-07-30
Feature/embedder

Dremars:
New feature embedder tool for creating embed html and links.

Need to check:
- ~~translations~~ done
- ~~area view handling~~ Moved to another card. Will be added later
- ~~possibly embed preview optimization relation to option changes.~~ Seems fine now

PR #462, opened 2020-09-04
New ToolMenu component

Dremars:
Added tool menu component separate from #429 so other tools can be moved forward
- New ToolMenu component
- New DropDownMenuButton component


Dremars:
[Työkaluvalikko](https://trello.com/c/h2RWQRJe/635-ty%C3%B6kaluvalikko)

mikkoisa (approved):
Mobile layout tool menu still missing, but most of the tools are probably not used/wanted on mobile so we need to decide that first.

PR #469, opened 2020-09-11
Check if unit has root service nodes

japauliina:
If unit does not have any services the API returns `root_service_nodes` as null. This broke the “Services and events” tab in the unit details.


Dremars (approved):
I'll merge this in a moment.

PR #470, opened 2020-09-11
Add other dynamic settings 

Arj-Peshtani:
Add dynamic settings for: 
- External translations
- Show/Hide area selection and nearby services
- Link to additional feedback info
- Show/Hide read speaker button in unit view

### DIFF
--- a/config/default.js
+++ b/config/default.js
@@ -78,6 +78,11 @@ if (typeof settings.SHOW_AREA_SELECTION === 'undefined') {
   settings.SHOW_AREA_SELECTION = true;
 }
 
+if (typeof settings.SHOW_READ_SPEAKER_BUTTON === 'undefined') {
+  // If not set default to Helsinki
+  settings.SHOW_READ_SPEAKER_BUTTON = true;
+}
+
 if (typeof settings.FEEDBACK_ADDITIONAL_INFO_LINK === 'undefined') {
   // If not set default to Helsinki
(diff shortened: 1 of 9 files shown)


Dremars (changes requested):
This doesn't seem feasible. I don't want us to start inserting individual translation into configuration file. We need to figure out some other way to insert translations. Maybe we can use configuration to add url for external translation JSON-file that is combined with our translation file? `const englishTranslations = { ...translations, ...overridingExternalTranslations } `  Or possibly just make the translation file configurable? 

I'm not entirely sure how this should be done but this doesn't seem to be it.

Arj-Peshtani:
I chose to follow your suggestion to add external json files for the translations. I have also added some other changes that need to be configurable for the city of Turku. Can you review it again, please? @Dremars

Dremars (changes requested):
@Arj-Peshtani Think we should avoid having city specific external translation files in git. I tried removing those files and using try-catch when requiring external translations. Seems to work fine except for a warning in build. So that could work.

Otherwise this looks good!

Arj-Peshtani:
@Dremars just removed the files and added try-catch when requiring external translations.

Dremars (approved):
Looks good. 

@mikkoisa Can you give a fast look and then lets merge this.

mikkoisa (approved):
Looks really good now, approved!

PR #486, opened 2020-09-24
Add support for loading accessibility sentences with PTV id

japauliina:
Helsinki’s accessibility sentences are loaded from a system with servicemap’s unit ids. Add support for loading the sentences from a system that uses PTV ids as identifiers.

Settings for using this API:
```
ACCESSIBILITY_SENTENCE_API: <URL>
USE_PTV_ACCESSIBILITY_API="true"
```


Dremars (approved):
@japauliina Code looks good. I assume there's no way to test how PTV accessibility sentences work without correct data?

japauliina:
@Dremars Unfortunately no, and since we do not have the updated version of the backend published yet it is hard to get the test data as well.

mikkoisa (approved):

PR #489, opened 2020-09-26
Fix/map marker handling

Dremars:
- Changing map marker handling to not remove search results when browsing results by selecting specific units. Selected unit is also highlighted by tooltip above cluster if marker is inside it.
- Changed new search to remove selected unit so it doesn't show tooltip on new results

### DIFF
--- a/client/client.js
+++ b/client/client.js
@@ -54,7 +54,6 @@ function Main() {
   // Remove server side styles
   React.useEffect(() => {
     const jssStyles = document.querySelector('#jss-server-side');
-    console.log(jssStyles);
     if (jssStyles) {
       jssStyles.parentElement.removeChild(jssStyles);
     }
--- a/src/components/SearchBar/SearchBar.js
+++ b/src/components/SearchBar/SearchBar.js
@@ -102,7 +102,7 @@ class SearchBar extends React.Component {
   }
 
   handleSubmit = (search) => {
-    const { isFetching } = this.props;
+    const { changeSelectedUnit, isFetching } = this.props;
     if (!isFetching && search && search !== '') {
       const {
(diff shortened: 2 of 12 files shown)


Dremars:
[(2) Haun löytämät toimipisteet näytetään kartalla kunnes käyttäjä tekee uuden haun tai siirtyy etusivulle](https://trello.com/c/7v1CMXQP/111-2-haun-l%C3%B6yt%C3%A4m%C3%A4t-toimipisteet-n%C3%A4ytet%C3%A4%C3%A4n-kartalla-kunnes-k%C3%A4ytt%C3%A4j%C3%A4-tekee-uuden-haun-tai-siirtyy-etusivulle)

mikkoisa (changes requested):
I really like this implementation! It is much easier to use the map when it is not zooming all the time. I noticed two bugs: Marker cluster hover seems to be broken (no hover popup at all). Also choosing a unit from search results, then going back to results page does not clear highlighted unit (tooltip stays on map).

mikkoisa:
We should also discuss the case where user selects unit from result page and the unit is not visible on the chosen zoom level/map position. Since we are no longer zooming on the unit, the user might have problems finding it on the map.

Dremars:
> I really like this implementation! It is much easier to use the map when it is not zooming all the time. I noticed two bugs: Marker cluster hover seems to be broken (no hover popup at all). Also choosing a unit from search results, then going back to results page does not clear highlighted unit (tooltip stays on map).

@mikkoisa 
Marker cluster hover has been activated once again (3e93589)
Hightlighted unit tooltip is hidden unless on unit page (2b2bf50)

mikkoisa (approved):

mikkoisa (approved):

PR #490, opened 2020-09-28
Show/Hide feedback permission dynamically

Arj-Peshtani:
Since in City of Turku feedbacks are published, this modification aims to show/hide the feedback permission dynamically.


Dremars (approved):

mikkoisa (approved):

Dremars:
@Arj-Peshtani I noticed that this just hides the field but the actual fetch is still sending `can_be_published` field as default `false` value. Is this desired functionality? Otherwise this seems fine.

Arj-Peshtani:
> @Arj-Peshtani I noticed that this just hides the field but the actual fetch is still sending `can_be_published` field as default `false` value. Is this desired functionality? Otherwise this seems fine.

We will be modifying the fields in the backend because except the "can_be_published" field we need to omit also some other fields when sending the request to the City of Turku feedback API.

Dremars:
@Arj-Peshtani Understood. Will merge this in a sec

PR #492, opened 2020-09-30
Fix/measuring tool line accessibility

mikkoisa:
- Make measuring tool line more accessible on high contrast map
- Add white background to measuring tool markers


Dremars:
Measuring markers and lines disappear on contrast change

Dremars:
@mikkoisa I'm getting "children with same key" error when changing contrast mode. 
Screenshot from 2020-10-01 10-40-23

Also after the error when changing measuring lines and changing back in contrast mode, the line changes are not applied. It seems to freeze into the state when error happened

Dremars (approved):

PR #493, opened 2020-10-01
Make municipality configurable and other setting related changes

japauliina:
Add a possibility for defining municipalities at setup. In case no data is given, default to capital area cities.

Also,
- Make URL to Reittiopas configurable.
- Fix FEEDBACK_IS_PUBLISHED setting.
- Add examples to env.example.
- Refactor config variable names to be consistent with JS naming conventions.


Dremars (approved):
Looks good. Good point with the naming conventions need to keep it more consistent in the future.

PR #497, opened 2020-10-07
Fix/mobile sticky tab

mikkoisa:
- Fixed TabList sticky calculations.
- Fixed issue with isMobile-hook in ExpandedSuggestions and Suggestion box: for some reason mobile check returned always false, thus causing sidebar overflow style to break on mobile view.
- Improved UnitView component to use isMobile hook instead of MobileComponent/DesktopComponent. Old implementation would briefly render non-mobile content (search bar etc.) when rendering UnitView.


Dremars (approved):
Seems to work well now.

PR #499, opened 2020-10-08
Prevent dialog showing error message when modalOpen changes to false

Dremars:
Feedback dialog showed error message when `modalOpen` changed from `send` to `false`. It seems like the dialog content updates before `open` value is changed to `false`. This caused flashing error message because text is changed in dialog but it did not close the dialog right away.

Changed dialog to not be rendered if `modalOpen` is `false`


Dremars:
[(1) Palautteen antamisen yhteydessä välähtää fronttivirhe](https://trello.com/c/bOvt2Lv2/666-1-palautteen-antamisen-yhteydess%C3%A4-v%C3%A4l%C3%A4ht%C3%A4%C3%A4-fronttivirhe)

mikkoisa (approved):

PR #501, opened 2020-10-14
Feature/printing

Dremars:
New printing tool for ToolMenu. Printing tool opens separate view on top of the current one with correct map state for printing purposes.

Updated Leaflet library


Dremars:
[(3) Tulostaminen](https://trello.com/c/YUTFfS0v/68-3-tulostaminen)

Dremars:
Need to add translations before merging

src/views/PrintView/PrintView.js
```diff
@@ -0,0 +1,365 @@
...
+                    const { name, street_address: address } = description.units[0];
+                    return (
+                      <React.Fragment key={description.number}>
+                        <StyledTableRow key={description.number}>
+                          <StyledTableCell>{description.number}</StyledTableCell>
+                          <StyledTableCell>
+                            {
+                              name
+                              && (
+                                <Typography variant="subtitle1">
```
mikkoisa, line 295: The unit name seems to be h6, which is confusing for screen readers. These could be all h1 or we could add a single main title h1 to the top of the page ("Printing view" for example) and these would be h2.

mikkoisa (changes requested):

src/views/PrintView/PrintView.js
Dremars: Those were meant to be paragraphs. It seems variant="subtitle1" change component to `h6`. Changing component to `p`

mikkoisa (approved):

PR #503, opened 2020-10-19
Fix highlight popup not showing when selected from list

Dremars:
Fixed popup not showing when selecting location from list rather than map. Need to figure out if popup should disappear when coming back from UnitView to SearchView

### DIFF
--- a/src/views/MapView/components/MarkerCluster/MarkerCluster.js
+++ b/src/views/MapView/components/MarkerCluster/MarkerCluster.js
@@ -41,6 +41,7 @@ const clusterData = {};
 
 const MarkerCluster = ({
   classes,
+  currentPage,
   data,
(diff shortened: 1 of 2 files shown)


Dremars:
[Toimipisteen valinta listalta ei korosta listan merkkiä](https://trello.com/c/XD8URmZ2/677-toimipisteen-valinta-listalta-ei-korosta-listan-merkki%C3%A4)

mikkoisa:
Looks good! Found small bug: when choosing a unit that is inside a cluster (either from result list or hovering cluster on map), the appearing popup is permanent, until user closes it or zooms out. However, this isn't really a problem, since popups can be closed manually.

mikkoisa (approved):

PR #504, opened 2020-10-20
Fix/map change error

mikkoisa:
- Fixed crash that happened when changing map to high contrast map and then opening unit from search results. This was caused by MapUtility not updating and using old map object.
- Fixed broken search view array compare that caused map to be centered to results every time props changed (when changing to high contrast)

### DIFF
--- a/src/.eslintrc
+++ b/src/.eslintrc
@@ -2,7 +2,8 @@
     "extends": ["airbnb"],
     "rules": {
         "react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }],
-        "no-shadow": 0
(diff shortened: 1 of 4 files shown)


mikkoisa:
[(1) Vahvakontrastisessa karttanäkymässä käyttäjä päätyy white screeniin](https://trello.com/c/L3LEO51s/669-1-vahvakontrastisessa-karttan%C3%A4kym%C3%A4ss%C3%A4-k%C3%A4ytt%C3%A4j%C3%A4-p%C3%A4%C3%A4tyy-white-screeniin)

Dremars (approved):

PR #506, opened 2020-10-23
Feature/area view geographical services

mikkoisa:
Geographical division services added to area view:
- Area tab has checkboxes for geographical divisions (postal code, neighborhood). These divisions can also be chosen by clicking on the map.
- Unit tab has different content when geographical divisions have been selected

Note: Embedding functionality comes in another PR.


Dremars (changes requested):
Noticed when selecting areas from Helsinki and then changing city settings to only vantaa: area's disappear from the map but units for selected Helsinki areas stay on map.

Dremars:
I noticed that focus highlight seems to hide texts. When using keyboard and selecting "geographical areas" new "loading" text is half way under focus highlight border/outline.  Also focus element for radio buttons seem to be partly on top of related labels.

Dremars (approved):
Seems to work well

PR #508, opened 2020-10-26
Feature/unit call information

mikkoisa:
Added call charge information to units with call charge that is not "pvm/mpm"
Example unit: Suomen löytötavarapalvelu

### DIFF
--- a/src/themes/index.js
+++ b/src/themes/index.js
@@ -51,6 +51,13 @@ const overrides = theme => ({
       padding: 12,
     },
   },
+  MuiAccordionSummary: {
+    root: {
+      '&$expanded': {
+        minHeight: 0,
+      },
+    },
+  },
 });
 
 const breakpoints = {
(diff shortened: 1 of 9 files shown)


src/themes/index.js
```diff
@@ -51,6 +51,13 @@ const overrides = theme => ({
       padding: 12,
     },
   },
+  MuiAccordionSummary: {
+    root: {
+      '&$expanded': {
+        minHeight: 0,
+      },
+    },
+  },
```
Dremars, line 60: Do we want this style change for all Accordions?  If it's just for specific component / view then we could just use styled component where we need it ```const StyledAccordionSummary = withStyles(theme => ({
  root: {
      '&$expanded': {
        minHeight: 0,
      },
  },
}))(AccordionSummary);```

mikkoisa: Good point! For now we want the style for all accordions, however that might change, so your suggestion is much better. Will change.

mikkoisa: Actually we should probably make ServiceMapAccordion component, since customized accordions are used on area view and soon on servicetree view. I think we should implement it on another branch after area view geographical services branch is merged to develop as well.

Dremars (approved):

PR #509, opened 2020-10-26
Add condition to hearing map fetch.

mikkoisa:
Fetch hearing maps only if unit has "kuuluvuuskartta" in keywords.

### DIFF
--- a/src/views/UnitView/UnitView.js
+++ b/src/views/UnitView/UnitView.js
@@ -78,9 +78,13 @@ const UnitView = (props) => {
     const unitId = params.unit;
     // If no selected unit data, or selected unit data is old, fetch new data
     if (!stateUnit || !checkCorrectUnit(stateUnit) || !stateUnit.complete) {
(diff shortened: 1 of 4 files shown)


mikkoisa:
[(1) Kuulokuvien hakemisen päivittäminen](https://trello.com/c/ypGw9YiU/685-1-kuulokuvien-hakemisen-p%C3%A4ivitt%C3%A4minen)

src/views/UnitView/UnitView.js
```diff
@@ -230,9 +234,9 @@ const UnitView = (props) => {
           {
             renderTitleForRS()
           }
-          {hearingMaps && (
+          {hearingMaps?.id === unit.id.toString(10) && (
```
Dremars, line 237: What does this do?

mikkoisa: This compares id of hearingMaps in redux state to current unit. Without this, it would render the hearing maps of the previous unit. The parameter inside toString seems to be unnecessary though.

Dremars (approved):

PR #510, opened 2020-10-29
Feature/news window

Dremars:
New AlertBox and NewsInfo components for showing announcements and error messages for users.


Dremars:
[(2) Uutisikkuna etusivulle uusista toiminnallisuuksista tai virhetilanteista](https://trello.com/c/jIfl6AQC/558-2-uutisikkuna-etusivulle-uusista-toiminnallisuuksista-tai-virhetilanteista)

mikkoisa (changes requested):
Looks really good! Noticed that while using screenreader, the news window content is not very clear. There could be a title that tells that it is a news window, because now the title says "Mittaa-etäisyys työkalu" which seems to imply that it is actually the tool itself. Also the close button does not say what it closes for the screen readers, so it could instead read something like "close news window" as sr-title.

mikkoisa:
Another thing to consider: when using screen reader jump link ("Jump to content") on top of page, the focus skips the news window. This probably works as it should, but can this cause screen reader users to miss important news?

Dremars:
Translations missing.

@mikkoisa Otherwise I've changed news alert and news box to have new heading texts. "Jump to content" is interesting problem. We could change it to be under title text but then it feels semantically out of place. First would be h2:"Home view"  then h2 or h3 "news info" etc. One thing could be changing "skip to content" link to focus into first heading in main area?

mikkoisa:
About the skip link: I think that the skip link should actually focus to the <main> tag if possible, since that marks the main content. Your suggestion to focus into first heading probably achieves the same thing.

mikkoisa (approved):

PR #512, opened 2020-11-02
Feature/area view embedder

mikkoisa:
- Added embedded view to area view.
- Fixed missing unit removal functionality from map view units. Now units can be toggled off in any view, if the option is available on embedder.


Dremars:
@mikkoisa Embeding swedish preschool education doesn't seem to work. Embed preview doesn't have any layers

Dremars (changes requested):
Previous comment

Dremars (approved):

PR #519, opened 2020-11-12
Feature/download tool dialog

Dremars:
New DownloadDialog for giving information and format selection before actually downloading data.


Dremars:
Waiting for translations

mikkoisa (changes requested):
Tested with Voiceover and works really well! However found issue when downloading data from service tree search results.

Näyttökuva 2020-11-16 kello 12 21 20
If we want to display service node name instead of id:
The names of the nodes can be found from service tree redux state (serviceTree.services) however, this will not work if search results page is the entry point, because then service tree has not been visited and serviceTree state is empty. In this case, I think service node names needs to be fetched on download tool dialog window.

mikkoisa (approved):
Looks really good and works well with voiceover.

PR #520, opened 2020-11-12
Focus geographical districts that are out of map

mikkoisa:
When choosing geographical district checkbox, focus district if it is out bounds.


mikkoisa:
Need to improve performance when multiple area checkboxes are chosen at the same time

Dremars (approved):
Seems to work well on production mode. There's some lagging when adding areas in geographical district.

PR #521, opened 2020-11-16
Feature/service tree units fetch

mikkoisa:
- Added units fetch to service tree view. Choosing a service node checkbox will fetch the node's units
- Modified service tree service node fetching. Now service tree will load all node child nodes when a node is expanded
- Added loading screen to map during service tree units fetch


src/views/ServiceTreeView/ServiceTreeView.js
```diff
@@ -111,47 +133,92 @@ const ServiceTreeView = (props) => {
...
 
+  const checkParents = (list, nodeID) => {
+    // This checks recursively if some parent node is included in given list
+    const serviceNode = services.find(service => service.id === nodeID);
+    if (!serviceNode || !serviceNode.parent) return false;
+    if (list.includes(serviceNode.parent)) {
+      return true;
+    }
+    return checkParents(list, serviceNode.parent);
+  };
```
Dremars, line 144: Wasn't this same code in serviceTree selector as well. Can we export that?

src/views/ServiceTreeView/ServiceTreeView.js
```diff
@@ -111,47 +133,92 @@ const ServiceTreeView = (props) => {
...
+        setTreeOpened([...opened, node.id]);
+        if (selected.includes(node.id) || checkParents(selected, node.id)) {
+          setTreeSelected([...selected, ...data.map(item => item.id)]);
+        }
+      })
+      .catch((e) => {
+        console.warn(e);
+        dispatchNodeFetch({ type: 'end', fetching: node.id });
+      });
+  };
```
Dremars, line 178: I'm wondering if this kind of redux logic should be in redux folder or as a seperate  utility file specific to ServiceTreeView. This component seems to have a lot of logic within.

Dremars (changes requested):
Spent some time to go through the code and it looks really good. Some minor improvements for clarity might be needed.

For accessibility I noticed that Loading texts when opening accordion element and when loading units are not read for NVDA. Also when unit loading is done there is no verbal indication that loading has completed. But the actual loading progress sound is working correctly which helps.

src/views/ServiceTreeView/ServiceTreeView.js
mikkoisa: Yes, it seems that this fetch does not actually interact directly with the component anymore, all fetch result data is passed through redux to the component. I'll move it to redux.

src/views/ServiceTreeView/ServiceTreeView.js
mikkoisa: Good call, will change.

mikkoisa:
Need to figure out how to handle selecting checkbox with screenreader. Loading units to map is useless to screenreader users.

PR #523, opened 2020-11-17
Fix TabLists out of bound index by defaulting to 0

Dremars:
There was a problem with `t` parameter trying to get index out of bound from given data array. This could happen when person searches results with multiple tabs, selects last tab, and changes city settings so that results don't give multiple tabs.


Dremars:
[(2) TabList noOrderer of undefined error](https://trello.com/c/qPMgTs7l/573-2-tablist-noorderer-of-undefined-error)

mikkoisa (approved):

PR #525, opened 2020-11-18
Change UnitGeometry to show MultiPolygon data

Dremars:
Draw MultiPolygon geometry for highlighted unit. Can be tested by searching "ulkoilusaaret" and selecting "Vasikkasaari..."

There's still a question if we want to show Polygons and Polylines in search for all units. But I think we can forward this one and add that later on if it's still required.


Dremars:
[(1) Ulkoilualueet tai Ulkoilusaaret eivät näy alueina](https://trello.com/c/bvnGUiL6/699-1-ulkoilualueet-tai-ulkoilusaaret-eiv%C3%A4t-n%C3%A4y-alueina)

mikkoisa (approved):

PR #527, opened 2020-11-20
Fix/area view usability

mikkoisa:
- Remove selected services (on geographical district services tab) when area selection changes
- If user has selected an address, automatically select address district checkbox when choosing a geographical district


mikkoisa:
[(3) Aluenäkymän parannukset](https://trello.com/c/NRVfCde9/703-3-aluen%C3%A4kym%C3%A4n-parannukset)

Dremars (approved):

PR #533, opened 2020-11-24
Add onclick event for unit polygon

Dremars:
Add onclick event for unit polygon.


mikkoisa (approved):

mikkoisa:
We could also add click event to unit polylines as well

Dremars:
> We could also add click event to unit polylines as well

@mikkoisa Good idea. Added onclick event for polylines as well

PR #541, opened 2020-12-08
Feature/nature areas

mikkoisa:
Added nature conservation areas to area view


Dremars:
Looks good. I'm wondering how this works with screen readers since this will not have any service locations to list. Maybe we can just hide nature area accordion from screen readers.

Dremars (approved):

PR #542, opened 2020-12-08
Fix/jest tests

Dremars:
- Fix ReadSpeaker interaction with SimpleListItem. After changing from aria-label to hidden DOM element in #534 hidden ReadSpeaker element is not necessary and causes multiple label readings for SimpleListItems.
- Fix jest tests
- Update snapshots


Dremars:
[Jest testien korjaus](https://trello.com/c/6fEVk95X/728-jest-testien-korjaus)

mikkoisa (approved):

PR #544, opened 2020-12-10
Use distance srText or default properly to empty

Dremars:
Distance object was by default an empty object making distance.srText to `undefined` rather than empty text


Dremars:
I'll add some jest tests so we can avoid this in the future

Dremars:
[(1) Toimipisteen-listaelementti sisältää tekstin undefined](https://trello.com/c/e7Xq0gH0/743-1-toimipisteen-listaelementti-sis%C3%A4lt%C3%A4%C3%A4-tekstin-undefined)

mikkoisa (approved):

PR #545, opened 2020-12-11
Feature/close sidebar

mikkoisa:
Added hide sidebar button


mikkoisa:
Translations still missing

Dremars:
When hiding sidebar and then adjusting screen width to mobile leaves user with a white screen.  This functionality should reset when changing from desktop to mobile or vice versa.

Dremars:
I noticed on NVDA I wasn't able to use keyboard to toggle sidebar. Seems like if HideSidebar button is aria-hidden it doesn't receive call any keyboard actions on NVDA. If I set aria-hidden to false it works with NVDA + keyboard.

Also noticed we have map div as aria-hidden=false which allows user to get into map element and starts reading all interactive elements on map (mainly leaflet controls)  And seems like map div's aria-label doesn't work anymore just like with service location's info items.

Dremars (changes requested):
Need to check accessibility. Look above comments

Dremars:
> I noticed on NVDA I wasn't able to use keyboard to toggle sidebar. Seems like if HideSidebar button is aria-hidden it doesn't receive call any keyboard actions on NVDA. If I set aria-hidden to false it works with NVDA + keyboard.
> 

Like we've now established (https://github.com/City-of-Helsinki/servicemap-ui/wiki/Accessibility#features-that-rely-on-sight-for-usability) This feature takes user to map which we've cropped out of scope for screen reader users. For this reason we should keep this element aria-hidden and inaccessible with screen readers.

mikkoisa:
Fixed issue with mobile version white screen.. Now sidebar hiding is completely disabled on mobile. Still, hiding sidebar on desktop and then changing to mobile and then back to desktop does not reset hidden state. However, I think this is such a rare case, that it is ok to work like this.

Dremars (approved):

PR #547, opened 2020-12-15
Fix/pagination accessibility

Dremars:
Fix pagination list to focus correctly to info element designed for screen readers. This helps both screen reader users and keyboard users to move into beginning of the list on page change.

Added jest tests for ResultList
Added browser test for service view's pagination focus handling


mikkoisa (changes requested):
Works really well except on iPhone with VoiceOver. I will try to find out what causes this.

mikkoisa:
Wasn't able to find the cause of the iOS Voiceover bug. When scroll position is on bottom, the focus to start of list does not work. Lets merge this, since the bug seems to be in VoiceOver itself.

mikkoisa (approved):

PR #548, opened 2020-12-17
Add new screen reader texts for AddressSearchBar when cleared and extra info

Dremars:


Dremars:
- Missing translations.
- Need to investigate focus problem with newer devices. Problem persist on Android v7 even with manual focus back to input. Need to check newer Androids and iPhone

src/components/AddressSearchBar/AddressSearchBar.js
```diff
@@ -93,7 +98,23 @@ const AddressSearchBar = ({
...
+    infoText = <FormattedMessage id="address.search.cleared" />;
+  } else {
+    infoText = (
+      <>
+        {infoText}
+        {
+          addressResults.length
+          && (
+            <FormattedMessage id="address.search.suggestion" />
+          )
```
mikkoisa, line 112: For some reason React displays "0" here if no address results. Works with:
` addressResults.length
            ? (
              <FormattedMessage id="address.search.suggestion" />
            ) : null`

mikkoisa (changes requested):

Dremars:
[(3) Kotiosoite-kentän korjaukset](https://trello.com/c/3fJ4BfEq/740-3-kotiosoite-kent%C3%A4n-korjaukset)

src/components/AddressSearchBar/AddressSearchBar.js
Dremars: Should be fixed now (bef2efa)

Dremars:
@mikkoisa Usability is still horrible on Android v7 but it seems to be inherent issue with that version. Can you check if all works fine on iPhone. I'll check Android v10 when I get a chance.

Dremars:
Seems to work pretty well on Android v10

mikkoisa (approved):
Works on iOS as well

Dremars:
Translations still missing.

PR #549, opened 2020-12-21
Fix/search view focus

Dremars:
Extended from #547 

Changed SearchView to use fixed PaginatedList component so we don't need pagination handling in two places.

Removed automatic focus on SearchView since it skipped all result ordering and tabs. Now when user comes to search page directly they'll start from the beginning of the page. If they come to search view from another page in Servicemap then focus is moved to back button like on all views. There might be a problem with user getting thrown behind search input. If user searched something on front page and now need to navigate through the same input-field to get to results. 

Not sure how this should work. Maybe focus to "x Hakutulosta" text automatically? But in that case finding back button becomes confusing.

### DIFF
--- a/src/components/Lists/PaginatedList/PaginatedList.js
+++ b/src/components/Lists/PaginatedList/PaginatedList.js
@@ -9,6 +9,7 @@ import PaginationComponent from '../../PaginationComponent';
 import { parseSearchParams, stringifySearchParams } from '../../../utils';
 
 const PaginatedList = ({
+  beforePagination,
   customComponent,
   data,
   id,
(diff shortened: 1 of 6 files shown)


Dremars:
[(1) Hakusivun automaattisen fokusoinnin korjaus](https://trello.com/c/jw9OsNq8/742-1-hakusivun-automaattisen-fokusoinnin-korjaus)

Dremars:
Missing translations and documentation update

mikkoisa:
This works really well now, the jump link to search results is a really good idea! I was thinking, could it be before the search bar? This way it would be even easier for screen reader users to navigate the page.

Dremars:
Good idea.

Dremars:
@mikkoisa I tried adding it but there seems to be problem in certain situation. For example when you do another search on search page focus doesn't change. In these cases jump link is behind.

mikkoisa (approved):

PR #551, opened 2020-12-22
Fix focus to ToolMenu when coming back from PrintView

Dremars:
- Fix focus to Toolmenu when coming back from PrintView
- Fix focus to ToolMenu when coming back from embedder
- New breadcrumb functionality with saved focusTarget data in order to refocus back to specific element when coming back with navigator

### DIFF
--- a/src/components/BackButton/BackButton.js
+++ b/src/components/BackButton/BackButton.js
@@ -24,7 +24,7 @@ const BackButton = (props) => {
   // Figure out correct translation id suffix
   let idSuffix = 'goToHome';
   if (breadcrumb.length) {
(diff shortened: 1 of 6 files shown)


Dremars:
[(2) Työkalujen fokusoinnin siirtäminen](https://trello.com/c/XcRUVXqq/735-2-ty%C3%B6kalujen-fokusoinnin-siirt%C3%A4minen)

Dremars:
I've removed history listen from Navigator in order to use breadcrumbs directly. I hope there's no hidden issues. While testing it seems to work just like before.

Added new breadcrumb functionality for refocusing when moving back in views

mikkoisa (approved):
Tested, and works well!

PR #557, opened 2020-12-31
Add back button to area view

mikkoisa:
Added title bar to area view. Decided to use "p" instead of heading, since the page title with the same name is just before title bar.


Dremars (approved):
Seems to work fine on Chrome, Firefox and Android 7

PR #558, opened 2021-01-04
Fix/mobile suggestion accessibility

mikkoisa:
- Suggestion arrow was not working at all on VoiceOver and TalkBack had issues with focus after pressing it. I suggest we disable it completely from screen readers, since it is a function that can be accessed later from search results. (Google does it like this as well)
- Added section-tag to tell screen readers where results start and end. Especially useful when browsing search history. Does not seem to work on TalkBack however.
- Disabled onBlur on mobile. On blur causes suggestion box to disappear when keyboard is closed on iOS. I think onBlur closing is not necessary on mobile?

### DIFF
--- a/src/components/ListItems/SuggestionItem/SuggestionItem.js
+++ b/src/components/ListItems/SuggestionItem/SuggestionItem.js
@@ -19,7 +19,6 @@ const SuggestionItem = (props) => {
     handleItemClick,
     handleArrowClick,
     icon,
-    intl,
     selected,
     subtitle,
     query,
@@ -53,13 +52,10 @@ const SuggestionItem = (props) => {
         onKeyDown={keyboardHandler(onClick, ['space', 'enter'])}
         onKeyUp={() => setMouseDown(false)}
         role="link"
-        tabIndex="0"
+        aria-label={`${text} ${subtitle || ''}`}
       >
         <span
           className={classes.container}
-          type="submit"
-          role="link"
-          tabIndex="-1"
         >
           <ListItemIcon aria-hidden className={`${classes.listIcon}`}>
             {icon}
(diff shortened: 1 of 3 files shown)


mikkoisa:
[Tarkenna painike mobiiliversion hakuehdotuksissa ei toimi ruudunlukijalla (VoiceOver)](https://trello.com/c/CST6vcWO/760-tarkenna-painike-mobiiliversion-hakuehdotuksissa-ei-toimi-ruudunlukijalla-voiceover)

Dremars:
@mikkoisa I think it's a good idea to disable Suggestion arrows on mobile screen readers. I'm still having massive problems using search field on Android v7 but then again seems like v7 just doesn't work. Android v7 didn't tell about section area either.

I'll test newer version later today

mikkoisa:
@Dremars If we want to look at fixing v7 issues, we could change it so that search bar "activates" only onClick instead of onFocus.

Dremars:
@mikkoisa I'm kinda lost what to do with this. v10 seems to work somewhat better but there's still problems getting back to input field from keyboard. Could be that I'm using TalkBack incorrectly. Does onclick activation cause problems on VoiceOver?

Also I still can't hear section being said anywhere on Android.

Dremars:
@mikkoisa Tested using onClick as activation but it seems that breaks activation after searching once

mikkoisa:
@Dremars Yeah onClick does not seem to work on this. I think we have to forget v7 and focus on v10. Closing the keyboard and getting back to input field is still an interesting issue. The only way seems to be by closing it from Android's bottom navigation buttons, but then getting back to input field is really difficult. However this seems to be an issue with Android itself, so nothing we can do about it.

PR #563, opened 2021-01-13
Fix/unit feedback accessibility

mikkoisa:
- Previously alert would trigger as soon as user entered page. Changed it so that it will activate only if user has visited feedback field
- Added focus to title on page mount


mikkoisa:
[(2) Toimipisteen palaute](https://trello.com/c/8AHXJHqO/745-2-toimipisteen-palaute)

Dremars:
Seems to work well.

Noticed couple other issues. When coming to unit's feedback view tile text is still "unit view" rather than "unit feedbackview" etc. This could be confusing since it says "unit view" and next element is back button with text "move back to unit view".

Also when coming back to unit view focus is lost on mobile.

Dremars (changes requested):
I'm having issue with back button on mobile unit feedback view. Back button doesn't actually take user back to unit view but to previous view before unit view.

Dremars:
Back button issue is old and will be fixed in another issue https://trello.com/c/YiUOH9l9/794-2-palauten%C3%A4kym%C3%A4-paluupainikke

Dremars (approved):

PR #571, opened 2021-01-21
Add skip links for suggestions box

Dremars:
Added skip links to before and after suggestion list.
This gives mobile screen reader user ability to close suggestions which was impossible before. This solution isn't perfect because user still can just move past these links leaving suggestions box over the current view.


Dremars:
This could be temporary solution for mobile search suggestion problem. I'm not sure if having 2 links is required we could also just have one link at the bottom of suggestions

mikkoisa (approved):
Tested and works well with iPhone VoiceOver.

PR #572, opened 2021-01-21
Fix external url to use getLocaleText function for translations

Dremars:
Fixed news info external url handling. External url was expecting string even though API returned translated object.


Dremars:
[(1) Uutisosion linkit eivät toimi](https://trello.com/c/H0ldvU4D/771-1-uutisosion-linkit-eiv%C3%A4t-toimi)

mikkoisa (approved):

PR #576, opened 2021-01-26
Fix/area accordion accessibility

mikkoisa:
Create accessible SMAccordion component and use it on service tree view and area view.


Dremars (changes requested):
I'm not able to access checkbox but if I can guess this is because checkbox just changes how map shows data? I think we need to add documentation about AreaView's accessibility decisions. Otherwise this might cause confusion for future developers.

Area view seem really heavy to use on mobile. Opening accordions takes ages to load and changing tabs seem to take long after opening some options.

mikkoisa:
Yes area view performance is really bad, especially on dev mode. I chose not to do major changes to it on this branch, since this is mostly about accordion accessibility. I think that the upcoming new version of area view would be a good opportunity to revisit these issues.

I chose to hide the unit tab checkboxes from screen readers because the functionality does nothing for blind users.

Dremars (approved):

PR #582, opened 2021-01-29
Set accessibility result order by default

mikkoisa:
- Use accessibility order by default if any accessibility settings are selected.


mikkoisa:
Not sure about the override importance. In these changes accessibility settings will override result orderer initialOrder prop. However, changing home address on address search bar will override accessibility ordering.

src/components/ResultOrderer/ResultOrderer.js
```diff
@@ -37,6 +38,14 @@ const ResultOrderer = ({
   setOrder,
   userLocation,
 }) => {
+  const userSettings = useSelector(state => state.settings);
+  const accessibiliySettingsLength = [
+    userSettings.mobility,
+    userSettings.colorblind,
+    userSettings.hearingAid,
+    userSettings.visuallyImpaired,
+  ].filter(i => (i !== false && i !== null)).length;
```
Dremars, line 47: I think we should move this to settings utility class as static function hasAccessibilitySettings or accessibilitySettingsActive etc.

Dremars (changes requested):

Dremars:
I think override order should be adress -> accessibility -> most relevant. Accessibility is setting which user selects for general use but address which is set by user for specific action or view.

src/components/ResultOrderer/ResultOrderer.js
mikkoisa: Implemented this on latest commits

Dremars (approved):

PR #588, opened 2021-02-02
Unit contract type municipality

mikkoisa:
Added municipality text next to contract type text.

### DIFF
--- a/src/components/ListItems/UnitItem/UnitItem.js
+++ b/src/components/ListItems/UnitItem/UnitItem.js
@@ -5,7 +5,7 @@ import UnitHelper from '../../../utils/unitHelper';
 import ResultItem from '../ResultItem';
 import SettingsUtility from '../../../utils/settings';
 import UnitIcon from '../../SMIcon/UnitIcon';
-import isClient, { uppercaseFirst } from '../../../utils';
+import isClient from '../../../utils';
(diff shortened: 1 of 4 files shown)


mikkoisa:
[(2) Kuntatieto sopimuslajin yhteyteen](https://trello.com/c/ivg765Oq/692-2-kuntatieto-sopimuslajin-yhteyteen)

Dremars (commented):
Seems to work mostly fine. Noticed how some searches like "laboratorio" doesn't show municipality text. They have municipality for unit object but department municipality is `null`

But I guess this is desired outcome?

mikkoisa:
@Dremars I think it is safer to display nothing than possibly incorrect information in that case. Municipality text is also hidden if we don't have translations for the city or if contract type is something else than municipal_service or purchase_service.

This got me thinking, should we change it to always show city name next to contract type if department municipality has value? Instead of just limiting it to those two contract types (municipal_service, purchase_service)

Dremars:
>This got me thinking, should we change it to always show city name next to contract type if department municipality has value? >Instead of just limiting it to those two contract types (municipal_service, purchase_service)

I'm not sure. Should propable ask from Mirjam.  But otherwise this looks good.

mikkoisa:
Decided that we will use department data instead of ID's to define if municipality should be shown.

Dremars (approved):

PR #589, opened 2021-02-03
New previousSearch functionality

Dremars:
New previousSearch functionality based on timestamps weighted with search count.

Change search times based solution to time based solution with weights from search counts. Weights are reduced every 5 days if user uses search.


Dremars:
[(2) Aikaisemmat haut. Missä järjestyksessä ovat?](https://trello.com/c/rRbeobXl/292-2-aikaisemmat-haut-miss%C3%A4-j%C3%A4rjestyksess%C3%A4-ovat)

mikkoisa (approved):
Looks really nice! It is hard to properly test at this point, but the logic seems good.

PR #590, opened 2021-02-04
Fix/mobile map improvements

mikkoisa:
- Added tap={false} to map to "fix" [safari double click problem](https://github.com/Leaflet/Leaflet/issues/7255). This option doesn't seem to break anything
- Changed mobile version marker click handling: now opens popup with link instead of navigating straight to unit page.

### DIFF
--- a/src/i18n/en.js
+++ b/src/i18n/en.js
@@ -280,6 +280,7 @@ const translations = {
 
   // Units
   'unit': 'Location',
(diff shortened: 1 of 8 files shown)


mikkoisa:
Missing translations

Dremars (changes requested):
@mikkoisa Mobile usability is way better now! Clusters click on close zoom levels seems to be broken. Click doesn't open usual cluster popup.

Dremars (approved):
Works nicely

PR #595, opened 2021-02-11
New ErrorBoundary component

Dremars:
New ErrorBoundary component for handling error rendering.

### DIFF
--- a/src/utils/user.js
+++ b/src/utils/user.js
@@ -4,6 +4,9 @@ class UserHelper {
   static useTheme() {
     return useSelector(state => state.user.theme);
   }
+  static useLocale() {
+    return useSelector(state => state.user.locale);
+  }
 }
 
 export default UserHelper;
--- a/src/utils/user.js
+++ b/src/utils/user.js
@@ -1,12 +1,4 @@
 import { useSelector } from 'react-redux'
-
-class UserHelper {
-  static useTheme() {
-    return useSelector(state => state.user.theme);
-  }
-  static useLocale() {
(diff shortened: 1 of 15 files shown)


Dremars:
Need to change:
- Figure out correct default message and get translations
- Default error styles

Dremars:
[(2) Virheviestisivu](https://trello.com/c/7z1pzNF5/838-2-virheviestisivu)

src/utils/user.js
```diff
@@ -0,0 +1,12 @@
...
+class UserHelper {
+  static useTheme() {
+    return useSelector(state => state.user.theme);
+  }
+  static useLocale() {
+    return useSelector(state => state.user.locale);
+  }
+}
+
+export default UserHelper;
```
mikkoisa, line 12: I think hooks can't / shouldn't be used in classes. This class can be replaced with:
```
import { useSelector } from 'react-redux'
 
export const useTheme = () => useSelector(state => state.user.theme);
export const useLocale = () => useSelector(state => state.user.locale);
```

mikkoisa: Otherwise looks and works really well!

mikkoisa (changes requested):

src/utils/user.js
mikkoisa: Ok, that makes sense! I was using the hooks plugin for eslint and it showed errors on that, however since it works I'm not sure if it should be error.

mikkoisa (approved):

PR #607, opened 2021-02-18
Add meta tags for units

Dremars:
Added required meta tags for creating cards for servicemap links in Twitter and Facebook


Dremars:
Facebooks works well now but Twitter is unable to fetch card info.

mikkoisa (approved):

PR #609, opened 2021-02-23
Add cookiehub code handling based on environment variables

Dremars:
Added cookiehub code based on environment variable. This is done because different cookiehub url is meant for specific domain but servicemap is used in 3 different domains currently


mikkoisa (approved):

Dremars:
[(2) CookieHub skriptin lisäys](https://trello.com/c/mBoGGwwq/822-2-cookiehub-skriptin-lis%C3%A4ys)

PR #616, opened 2021-03-05
Feature/show selected settings on topbar

Dremars:
- Refactored DrawerMenu as subcomponent for TobBar component since it's only used there. 
- New SettingsText component that renders specific settings text  showing all selected settings under a title.
- Updated TopBar and DrawerMenu to show selected settings under settings button

Missing translations and need to test accessibility with different devices


Dremars:
Waiting translations

mikkoisa (approved):
Looks good!

PR #617, opened 2021-03-05
New PanControl component for panning leaflet map

Dremars:
New PanControl component for handling map panning with keyboard.

~Still need to add direct keyboard handling for moving map with arrow keys.~ Added in 058ef1e

### DIFF
--- a/src/views/MapView/MapView.js
+++ b/src/views/MapView/MapView.js
@@ -192,6 +192,15 @@ const MapView = (props) => {
     setTimeout(() => {
       const e = document.querySelector('.leaflet-control-zoom');
       const e2 = document.querySelector('.leaflet-control-attribution');
+      // Adjust bottom controls to align with pan controls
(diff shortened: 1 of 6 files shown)


Dremars:
[(3) Kartan navigointipainikkeet](https://trello.com/c/Uz9E5IkD/560-3-kartan-navigointipainikkeet)

mikkoisa (commented):
Looks good! Should we consider centering the map buttons by adding margin to the other buttons?  
Näyttökuva 2021-3-10 kello 8 10 41

Dremars:
Fixed control stylings 9756419 and fixed issue with leaflet maptype change causing adjusted styles to disappear 577ceab

mikkoisa (approved):
Looks good and works well! We could also at some pint add similar key press handlers for "+" and "-" keys that would zoom the map.

PR #620, opened 2021-03-10
Fix/multiple popups

mikkoisa:
- Fix multiple popup problem when choosing different units from marker cluster list.
- Move marker cluster unit list click from span to li. Previously click event would not trigger when clicking the edge of the list item


mikkoisa:
[(1) Kartan popupit jää roikkumaan kartalle](https://trello.com/c/QtaExrMV/835-1-kartan-popupit-j%C3%A4%C3%A4-roikkumaan-kartalle)

Dremars:
I can't see the difference in cluster unit selection's  popup functionality between this branch and production. It's still possible to open multiple popups from different clusters and I'm not able to open multiple popups from same cluster.

Dremars (approved):

PR #621, opened 2021-03-11
Feature/new area view

mikkoisa:
~Translations missing~


Dremars:
New area view seems to work really well now and feels way better than before. Usability is amazing and aria-live elements work really well and inform user with all the necessary info. Most of my accessibility issues are related to problems of understanding difference between sections. I tried to use new area view blindly but got lost where I was in relation to categories, subcategories, and service location lists.

### Accessibility - Mobile
#### Major issues
- Map button should be aria-hidden
- I feel like categories and sub categories get mixed with each other. It's hard to know when sub categories end and new category starts. Maybe we could give heading levels or make subcategories work like lists where reader tells user when list ends. Maybe headings for main categories would be better  because list can become weird with so much inner content (service location listings etc.)
- Service locations could also be a list so user get indicated where they are progressing in list and also when the list ends.
- City areas and postal codes: Functionality is hard to understand. First user selects "kaupunginosa" then they get offered list of services in area and after that it's straight to "Helsinki". I think it would help if we had some kind of title or descriptive text here that points how the next checkbox list is used for selecting filter attributes for services. Maybe "Select city areas to inspect" (etc.) then it's way more understandable what Helsinki means in this context.

#### Minor issues
- Would be great if we can change address search suggestions to be read as list so user knows when list ends and starts. Like with search suggestions
- I wonder if for city area and postal code section's service location listing should say more than just number? Now it's "Asiakaspalvelu (vesihuolto) (1)" but would it be better to have aria-label etc. say "Asiakaspalvelu (vesihuolto) (1) toimipiste"

Dremars (approved):

PR #623, opened 2021-03-12
Add timeout to search

mikkoisa:


src/redux/actions/unit.js
```diff
@@ -11,10 +12,18 @@ const {
...
-  abortController = null,
 ) => async (dispatch, getState) => {
   const { units, user } = getState();
   const { locale } = user;
+  const { searchTimeout } = config;
+
+  const timeout = searchTimeout;
+  const abortController = new AbortController();
+  const fetchTimeout = setTimeout(() => {
+    console.warn('Search backend not responding');
```
Dremars, line 23: Can we change this to "Search fetch aborted: Timeout after ${searchTimeout/1000} seconds"

Dremars (commented):
I think this could be used for all fetches and add it to our fetch functionality. If its big change we can go with this for now. I would like to improve our whole fetch functionality in the future.

src/redux/actions/unit.js
mikkoisa: Yes! That text would be much better.

Dremars (approved):

PR #628, opened 2021-03-25
Fix/settings button errors

mikkoisa:
Fixed several errors that caused warnings on console


Dremars (approved):
Good catch. Somehow missed those when changing translation IDs

PR #632, opened 2021-03-29
Fix/optimize rerendering

mikkoisa:
- Changed all components that used old getLocaleText to function components so that new useLocaleText hook can be used.
- Created new useNavigationParams hook to replace old getAddressNavigatorParams that caused re-render even time any redux state changed. 
- Changed all components that used this to function components so that new useNavigationParams hook can be used 

note: search view was excluded from this and should be converted to function component when search functionality will be updated: https://trello.com/c/wX0TgyCV/701-2-hakutulosten-määrä-kartalla


Dremars:
Looks good! Pretty large PR so might have missed something but seemed fine

Dremars (approved):

PR #636, opened 2021-03-31
Add sitemap

mikkoisa:
- Create sitemap on server start and once a week
- Removed embed views from search robots

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,6 +7,7 @@ MODE=development
 # AppKey for appdynamics browser tracking
 # APP_DYNAMICS_APP_KEY=
 
+DOMAIN="https://palvelukartta.hel.fi"
 ACCESSIBILITY_SENTENCE_API="https://www.hel.fi/palvelukarttaws/rest/v4"
 SERVICEMAP_API="https://api.hel.fi/servicemap/v2"
 EVENTS_API="https://api.hel.fi/linkedevents/v1"
--- a/config/default.js
+++ b/config/default.js
@@ -171,6 +171,7 @@ export default {
     "shortcomings": "#b00021",
   },
   "production": settings.MODE === 'production',
+  "domain": settings.DOMAIN,
   "initialMapPosition": settings.INITIAL_MAP_POSITION.split(','),
   "servicemapURL": settings.SERVICE_MAP_URL,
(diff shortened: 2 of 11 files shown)


mikkoisa:
Robots.txt file might be a little underutilised here. For now it is used only to store sitemap location.

mikkoisa:
[(3) Hakukoneoptimointi: sitemap](https://trello.com/c/FKgfRCHY/842-3-hakukoneoptimointi-sitemap)

mikkoisa:
[(1) Hakuoneoptimointi: piilotetaan tietyt sivut hakukoneilta](https://trello.com/c/fR4AmxZD/841-1-hakuoneoptimointi-piilotetaan-tietyt-sivut-hakukoneilta)

Dremars:
I'm not sure if this works correctly. I'm not able to load it but looks like it's single text and inspecting looks like tags are `<url />` tags. Watching some other sites they seem to use `<sitemap />` and `<loc />` tags. https://www.oracle.com/sitemap.xml and https://www.google.com/sitemap.xml To be perfectly honest I'm not exactly sure how this should show but I think we should have somewhat similar file as others

mikkoisa:
@Dremars Those examples seem to be sitemap index files, which list all of the different sitemaps. If you open one sitemap, it looks similar to ours. For example: https://www.google.com/gmail/sitemap.xml. We need only one sitemap, so sitemap index is not necessary

Dremars:
@mikkoisa Good point didn't notice that. I guess this is good then. One thing that slightly concerns still is that opening sitemap.xml keeps my cpu running for as long as I have that page open. Could be just my computer or the size of that file. But even after page is loaded it still keeps processing something

Dremars:
@mikkoisa Now it worked fine. Weird

Dremars (approved):

mikkoisa:
@Dremars I am also unfamiliar with sitemaps, so we have to see how this works on server. If we encounter performance issues, we can split the sitemap to smaller sitemaps and add index file (like in those examples you provided). However our sitemap size is within limits defined by google documentation https://developers.google.com/search/docs/advanced/sitemaps/build-sitemap#general-guidelines

PR #638, opened 2021-04-01
Check that description has units

Dremars:
Only add markers that are unit markers or clusters. 

Print view was crashing because it was fetching any leaflet marker including transit stops. Changed marker fetching to only get markers that have icons with class `unitMarker` or `unitClusterMarker`. Also added check that confirms description has existing first unit.


Dremars:
[(1) Tulostus ei toimi](https://trello.com/c/KWZYcX0F/858-1-tulostus-ei-toimi)

mikkoisa (approved):

PR #640, opened 2021-04-13
Fix/hide button color

mikkoisa:
Fix issue with bottom right map button styles affecting hide sidebar button styles.
Use translations on button.


Dremars (approved):
Missing swedish translations. Otherwise looking good

mikkoisa:
@Dremars Let's add missing translations later.

PR #643, opened 2021-04-15
Feature/entrances

mikkoisa:
Added unit's additional entrances to map and sidebar. Example unit with entrances: 8215.


src/components/SMAccordion/SMAccordion.js
```diff
@@ -11,20 +11,24 @@ const SMAccordion = ({
   collapseContent,
   onOpen,
   disabled,
+  disableUnmount, // Disables accordion collapse content unmount on accordion close
```
Dremars, line 14: What's the point of this? Does this keep Accordion content rendered? Was there reason this was needed?

Dremars (commented):
Looks good. One thing could be checked. On higher zoom levels entrances are covering unit marker. Maybe we could change entrances to show only on certain zoom levels?
entrances

src/components/SMAccordion/SMAccordion.js
mikkoisa: Unmounting accordion content on close is unnecessary and makes the closing animation clunkier. Also not unmounting makes re-opening smoother, since no mounting is required. The MUI-accordion component actually works like this by default. However this functionality is bad for long accordion content (like accordions in service list view and area view), since it affects performance in those cases, so for that reason this prop hasn't been needed before.

mikkoisa: Good idea about the entrances showing only on closer zoom levels! Will fix it

Dremars: Sounds good.

Dremars (approved):

PR #644, opened 2021-04-16
Feature/search results functionality

mikkoisa:
- Choosing search result removes other units from map and zooms closer
- Map position is saved when selecting search result. When returning to results page, map view is returned to this position
- Updated marker cluster icon styles


src/views/UnitView/UnitView.js
```diff
@@ -60,11 +60,13 @@ const UnitView = (props) => {
   const checkCorrectUnit = unit => unit && unit.id === parseInt(match.params.unit, 10);
 
   const [unit, setUnit] = useState(checkCorrectUnit(stateUnit) ? stateUnit : null);
+  const viewPosition = useRef(null);
```
Dremars, line 63: Isn't ref used for DOM nodes? Is there a reason ref is used here instead of state?

mikkoisa: I used useRef because component unmount function on useEffect can't use state (it does not receive state updates). Ref persists trough whole component lifecycle. React docs mention that useRef can actually be used for more than Dom nodes nowadays: _However, useRef() is useful for more than the ref attribute. It’s handy for keeping any mutable value around similar to how you’d use instance fields in classes_ https://reactjs.org/docs/hooks-reference.html#useref

Dremars (approved):

PR #645, opened 2021-04-16
Feature/new title bar

mikkoisa:
New style for title bars with back button


Dremars (approved):
Looks ok. If screen readers have been tested we can merge

PR #648, opened 2021-04-22
Feature/area suggestions

mikkoisa:
* Added areas to search suggestions
* Refactored area view mount url parameter handling
* Fixed area view embedding problems


Dremars:
Noticed that if you have something opened in area view when you select another area category from search then correct area is not opened in area view

Also education areas don't seem to open any data on map but I guess this is inteded since we don't know which semester they want to inspect?

mhieta (approved):
Works as expected now 👍

PR #654, opened 2021-04-29
Top bar settings button roles

mikkoisa:
Accessibility report section 4.7
- Changed role aria-pressed to aria-haspopup
- Added jest tests


src/components/TopBar/SettingsButton/__tests__/SettingsButton.test.js
```diff
@@ -0,0 +1,79 @@
...
+};
+
+
+describe('<SettingsButton />', () => {
+  // let render;
+  let mount;
+
+  beforeEach(() => {
+    // render = createRender({ wrappingComponent: Providers });
+    mount = createMount({ wrappingComponent: Providers });
```
Dremars, line 58: I think we could use shallow rendering in this test case. It makes snapshot way shorter and I think it allows seeing roles without problem. Mount is usually needed if component does some logic on mount or I think click events etc require mount as well. But this seems like it could work with shallow

Dremars (changes requested):

src/components/TopBar/SettingsButton/__tests__/SettingsButton.test.js
mikkoisa: Good point! Lets change it

Dremars (approved):

PR #656, opened 2021-04-30
Add title bar to service tree view

mikkoisa:
Accessibility report 4.10


Dremars (commented):
I noticed how `h2#view-title` for serviceTreeView is "Palveluluettelo" and `h3` is "Palveluluettelo" as well. I was wondering if we should change view-title for service tree to "Palveluluettelosivu" or something similar? But I guess there's no technical problem. Most view titles are pages anyway.

Otherwise looks good!

PR #662, opened 2021-05-07
Filter duplicate feedback links

mikkoisa:
From accessibility report 4.4


mikkoisa:
@Dremars Most of Vantaa city units, for example https://palvelukartta.hel.fi/fi/unit/18465

Dremars (approved):

PR #670, opened 2021-05-11
Remove transit popups from screen readers

mikkoisa:
From accessibility report 3.1

### DIFF
--- a/src/views/MapView/components/TransitStops/TransitStopInfo/TransitStopInfo.js
+++ b/src/views/MapView/components/TransitStops/TransitStopInfo/TransitStopInfo.js
@@ -93,7 +93,7 @@ const TransitStopInfo = ({ stop, onCloseClick, classes }) => {
 
 
   return (
-    <div className={classes.tranistInfoContainer}>
+    <div aria-hidden className={classes.tranistInfoContainer}>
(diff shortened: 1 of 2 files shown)


Dremars (changes requested):
This doesn't seem to fix problem. Transit stop markers themselves are read as numbers. This is probably because they have following styles `.icon-icon-hsl-bus:before {
    content: "2"; }`
Which I guess is used to figure out which icon to use? But these markers should be `aria-hidden` or they are read as "1", "2" etc.

Dremars:
There are also some empty elements in bottom right controls. I'm not sure which ones are causing that

mikkoisa:
@Dremars I see, didn't realise that the numbers were visible to screen readers. Fixed the problem

Dremars (changes requested):
Transit icons are still read as numbers. Problem is that icon itself is aria-hidden but transit icons have 2 spans one is `transitBackground` which is not aria-hidden this should be changed to aria-hidden as well. It's in getTransitIcon function.

mikkoisa:
@Dremars Fixed the background span too. Safari + Voiceover didn't have this issue, so I didn't notice it at all.

Dremars (approved):
@mikkoisa Interesting how differently all these work. :smile: But now it works perfectly with NVDA as well

PR #673, opened 2021-05-17
Fix/map button focus

mikkoisa:
- Move zoom buttons to PanControls compoent
- Make UserLocation button aria-hidden. Update button margin adjusting
- Add default outline to links

### DIFF
--- a/browserTests/addressTest.js
+++ b/browserTests/addressTest.js
@@ -34,7 +34,7 @@ test('AddressView does render correct view', async (t) => {
   ;
 });
 
-test.only('AddressView map renders correctly', async (t) => {
+test('AddressView map renders correctly', async (t) => {
   const zoomOut = await Selector('.zoomOut');
   await t
     .click(zoomOut)
--- a/browserTests/addressTest.js
+++ b/browserTests/addressTest.js
@@ -34,8 +34,8 @@ test('AddressView does render correct view', async (t) => {
   ;
 });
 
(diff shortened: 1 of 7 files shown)


browserTests/addressTest.js
```diff
@@ -34,8 +34,8 @@ test('AddressView does render correct view', async (t) => {
   ;
 });
 
-test('AddressView map renders correctly', async (t) => {
-  const zoomOut = await Selector('.leaflet-control-zoom-out');
+test.only('AddressView map renders correctly', async (t) => {
```
Dremars, line 37: I guess this .only was for testing purposes? Should be removed

browserTests/browserTest.js
```diff
@@ -67,8 +67,8 @@ fixture`Map tests`
     await waitForReact();
   });
 
-test('Transit marker visible after zoom', async (t) => {
-  const zoomIn  = Selector('.leaflet-control-zoom-in');
+test.only('Transit marker visible after zoom', async (t) => {
```
Dremars, line 70: I guess this .only was for testing purposes? Should be removed

Dremars (changes requested):

browserTests/addressTest.js
mikkoisa: Good catch! Will remove

Dremars (approved):
Looks good and this is working amazingly right now! Really like the keyboard usability

PR #676, opened 2021-05-21
Use correct locale parameter on embedder text

mikkoisa:


src/views/EmbedderView/EmbedderView.js
```diff
@@ -79,6 +79,7 @@ const EmbedderView = ({
   const selectedUnit = useSelector(state => state.selectedUnit.unit.data);
   const currentService = useSelector(state => state.service.current);
   const getLocaleText = useLocaleText();
+  const userLocale = useSelector(state => state.user.locale);
```
Dremars, line 82: Lets try to unify these selectors to hook functions for easier changes in future. We have `useUserLocale` hook in `src/utils/user.js`

Dremars (changes requested):

Dremars (approved):

PR #679, opened 2021-05-24
Disable crawling with robots file if on staging

mikkoisa:


server/sitemapMiddlewares.js
```diff
@@ -20,6 +20,21 @@ export const getSitemap = (req, res) => {
   }
 }
 
+export const getRobotsFile = (req, res, next) => {
+  // Returns robots.txt file that points to sitemap location
+  if (config.domain) {
+    res.type('text/plain');
+    if (config.serverType === 'staging') {
```
Dremars, line 27: Since this is only on server can't we use `process.env.SERVER_TYPE`. This would also removes the need for passing this data to client in HTML portion of server file and extracting it to config file. Those 2 could be removed

Dremars (changes requested):

server/sitemapMiddlewares.js
mikkoisa: Good idea!

Dremars (approved):

PR #681, opened 2021-05-28
Feature/accessibility info on unit view

mikkoisa:
- Add accessibility settings info to unit view accessibility tab
- Update accessibility info styles
- Fix server mismatch error on description text


mhieta (approved):
Works as expected 👍

PR #682, opened 2021-05-31
New HTTPClient and ServiceMapAPI classes for imporved fetch handling

Dremars:
New HttpClient and ServiceMapAPI classes for better API fetching.

HttpClient has in-built abort handling and next fetching logic.
ServiceMapAPI extends HttpClient and contains all required fetch data in individual methods that better describe what is being fetched from Servicemap API.

This way we can get rid of the constants file and messy fetch logic used in previous solution.


mikkoisa (approved):
Looks really good! 👍

PR #693, opened 2021-06-14
Add map centering to event view

mikkoisa:
Previously event view would not center map to unit location if arriving straight to event view. Example event url: palvelukartta.hel.fi/fi/event/helmet:220550


mhieta (approved):
Good, seems to work as expected 👍

PR #694, opened 2021-06-15
Fix/duplicate area data

mikkoisa:
Problem was that fetching school area types of certain school year updated all years with the new data. For example school year 2021-2022 also got year 2020-2021 data, causing duplicate data.

Also removed old school year data


mhieta (approved):
Looks good, no issues

PR #695, opened 2021-06-15
Fix/embedded nature popup

mikkoisa:
- Fix embedded nature area view missing popup
- Ad  left side map focus border to views without sidebar

### DIFF
--- a/src/views/MapView/MapView.js
+++ b/src/views/MapView/MapView.js
@@ -137,6 +137,7 @@ const MapView = (props) => {
   };
 
   const setClickCoordinates = (ev) => {
+    if (embeded) return;
     setMapClickPoint(null);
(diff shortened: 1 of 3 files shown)


src/views/MapView/MapView.js
```diff
@@ -137,6 +137,7 @@ const MapView = (props) => {
   };
 
   const setClickCoordinates = (ev) => {
+    if (embeded) return;
```
mhieta, line 140: Should this constrant be `embedded ` ?

mhieta (changes requested):
Please check/fix constant naming, otherwise good 👍

src/views/MapView/MapView.js
mikkoisa: True, will change those!

mhieta (approved):
Good 👍

PR #697, opened 2021-06-16
Area view unit list amount

mikkoisa:
- Added overlapping district units to list of units
- Updated wording: services -> units


src/views/AreaView/components/DistrictUnitList/DistrictUnitList.js
```diff
@@ -87,7 +87,19 @@ const DistrictUnitList = (props) => {
         </div>
       );
     }
-    const districtsWithUnits = district.data.filter(obj => obj.unit);
+    const districtsWithUnits = [];
+    district.data.forEach((obj) => {
+      if (obj.unit) {
+        districtsWithUnits.push(obj);
+      }
+      if (obj.overlaping) {
```
mhieta, line 95: If this `overlaping` -attrbute is defined by us, then it should be changed to `overlapping`

mhieta (changes requested):
@mikkoisa Please update attribute-typos.

src/i18n/en.js
```diff
@@ -47,8 +47,8 @@ const translations = {
   'area.tab.publicServices': 'Public service areas',
   'area.tab.geographical': 'Neighbourhoods and postal code areas',
   'area.services.local': 'Services in your own area',
-  'area.services.nearby': 'Services in nearby areas as a list',
-  'area.services.all': 'Services as a list',
+  'area.services.nearby': 'Units in nearby areas as a list',
+  'area.services.all': 'Units as a list',
```
Dremars, line 51: Unit should be Service location(s)

Dremars (changes requested):

src/views/AreaView/components/DistrictUnitList/DistrictUnitList.js
mikkoisa: Fixed 👍

src/i18n/en.js
mikkoisa: Fixed 👍

mhieta (approved):
Looks good now 👍

Dremars (approved):

PR #700, opened 2021-06-18
Fix/rescue district data

mikkoisa:
Adds missing rescue districts to area view.


Dremars (approved):
Looks OK and seems to work fine

PR #701, opened 2021-06-18
Prevent automatic focus to site title if cookiehub banner is not hidden

Dremars:
Check if cookiehub banner is hidden before focusing to site title


mikkoisa (approved):
Looks good 👌

mhieta (approved):
Good 👍

PR #708, opened 2021-07-15
Prevent vertical overflow being hidden when not in mobile

Dremars:
There was an issue with suggestion box overflowing out of sidebar wrapper when wrapper has `overflow hidden`. This was caused by suggestion box setting overflow hidden. For solution I changed overflow hidden only to appear in mobile widths.

Suggestion box also had horizontal overflow scroll being shown. This is caused by list item dividers having margins that take it out of parent bounds. I've added `overflow-x: hidden` for suggestion container.

Removed `max-height` from suggestion area because it doesn't seem to do anything because it is not responsive to different heights.

Changed searchbar wrapper to relative position so suggestion area can be set width 100% based on wrapper. This saves us from having to worry about having correct `left` and `right` attributes for suggestion area because width will be always in relation to wrapper.

### DIFF
--- a/src/components/SearchBar/SearchBar.js
+++ b/src/components/SearchBar/SearchBar.js
@@ -347,7 +347,7 @@ class SearchBar extends React.Component {
     }  ${
       className
     }`;
-    const wrapperClasses = `${isActive ? classes.mobileWrapper : classes.wrapper}`;
+    const wrapperClasses = `${isActive ? classes.mobileWrapperActive : classes.mobileWrapper}`;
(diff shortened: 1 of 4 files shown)


Dremars:
[(1) Hakuhistoria ei scrollattava](https://trello.com/c/jwK77y2k/942-1-hakuhistoria-ei-scrollattava)

mikkoisa (approved):
Looks good! Tested with mobile device and everything seems to work as it should

PR #710, opened 2021-07-23
Create sitemap only when SERVER_TYPE is production

Dremars:
Rather than only disabling sitemap when SERVER_TYPE is staging change sitemap to be disabled by default. Changed to sitemap creation to check if SERVER_TYPE is production and NODE_ENV is production. This makes sure sitemap is not created in development mode or in staging server.

This requires changing environment variables in production server. Need to add SERVER_TYPE=production into envrionment variables


mikkoisa (approved):
Good solution!

PR #711, opened 2021-07-26
Feature/event search page

mikkoisa:
- Added events to search suggestions and results. 
- Replaced `fetchUnits` with new `fetchSearchResults` fetch action that uses the new fetch functionality.
- Replaced `units` state from redux with `searchResults` 
- Changed tab styles, including scrollable mobile tabs (based on designs linked to the trello card) 

(good search query to test new features is once again "ranta")

NOTE: this PR does not include unit view events changes (as seen on the Trello card designs)


mikkoisa:
[(5) Tapahtumien haku LinkedEventsin kautta](https://trello.com/c/L7WZpy95/321-5-tapahtumien-haku-linkedeventsin-kautta)

mikkoisa:
Noticed a bug that prevents service tree search, needs to be fixed

src/redux/actions/search.js
```diff
@@ -0,0 +1,78 @@
...
+
+  const onNext = (total, max) => {
+    dispatch(fetchProgressUpdate(total, max));
+  };
+
+  if (options.q) {
+    smAPI.setOnNext(onNext);
+    results = smAPI.search(options.q);
+  } else if (options.service_node) {
+    smAPI.setOnNext(onNext);
```
Dremars, line 29: Does `onNext` need to be set in if else statements or can it be outside? Does something bad happen if options doesn't include `q` or `service_node` and `onNext` is set?

Dremars (commented):
Code looks good. 

Some questions about tablist color on mobile being different blue than other content. Also tab buttons seem to be smaller with less padding which has it's pros and cons. This means we'll lose some usability by buttons being harder to press and more prone to miss clicks. But it gives more room to other content which is nice in mobile. But if these have already been considered in designs then we can proceed with those.

Good improvement and nice to have more functionalities moved to use new fetch

PR #713, opened 2021-08-17
update user position constantly

mikkoisa:


Dremars (commented):
Seems to work fine but I'm concerned about the usability and how understandable search distances are. I think we should have location marker separated from user marker. (See HSL route map) Now it's slightly confusing as to what these distances are in relation of since the location marker is moving. It makes sense when using address search to move location marker to specific position but when location marker is updated distances don't mean anything.

Should talk about this with designer.

PR #715, opened 2021-08-24
Update Webpack and React Leaflet

mikkoisa:


Dremars (changes requested):
I noticed that it is possible to click through pan controls to map. These events shouldn't be propagated or bubble outside of the clicked element. For some reason new map click event handling doesn't take into account the actually clicked event and causes AddressPopup to show when attempting to click controls.

Dremars (approved):

Dremars (approved):

PR #718, opened 2021-08-26
Feature/update unit services tab

mikkoisa:
- Updated unit view services tab and service, event, and reservation lists.
- Updated SMButton, TitledList and some list item components.
- Added new styles to unit popups
- Added new marker icon for units with events (only on unit view). Also added closest event to unit's popup.

~~Note: still missing some translations~~


Dremars (approved):
Missing translations but otherwise looks OK and seems to work fine.

PR #723, opened 2021-09-03
Fix/new ortographic map

mikkoisa:
Enable ortographic map that has been disabled.

- Use WMS service to render the map instead of WMTS. HSY's WMTS data doesn't work on our map.
- Updated ortographic map WMTS url, even though it is not used now


mikkoisa:
Need to merge React Leaflet update PR before merging this.

Dremars (approved):

PR #725, opened 2021-09-03
Feature/linked accessibility settings

Dremars:
There was a need to be able to link UnitView page to others with accessibility info showing same info what the person creating the link sees. For this purpose we needed to get accessibility setting sent to receiving user. For this purpose we have 2 new dialogs: first dialog is created for helping user to create an actual link with search parameters, second one is for accepting settings if settings exist as search parameters and saves these into user's settings if accepted.

New components:
- AcceptSettingsDialog - for accepting settings given in url
- LinkSettingsDialog - for helping user to create a link with accessibility settings


mikkoisa (approved):
Looks good! We should use that nice text copy field on embedder tool as well.

PR #727, opened 2021-09-07
Update test libraries. Add quarantine mode

mikkoisa:
[Use quarantine mode on tests](https://testcafe.io/documentation/402830/guides/basic-guides/run-tests#quarantine-mode). This mode repeats failing test n times if it fails. This should help with unstable tests that fail for no reason and need to be restarted. Quarantine mode configured to attempt failing tests 5 times and to require at least 2 successes to approve.

Updated test libraries for this feature.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7598,9 +7598,9 @@
           "dev": true
         },
         "ws": {
-          "version": "7.5.3",
-          "resolved": "https://registry.npmjs.org/ws/-/ws-7.5.3.tgz",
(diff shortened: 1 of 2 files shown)


mhieta (changes requested):
@mikkoisa Looks good, but could you update values a bit:
```
attemptLimit=5,successThreshold=2
````

Dremars (approved):

mhieta (approved):
Looks good now. Let's evaluate this in practice and change the values if necessary.

PR #728, opened 2021-09-07
Use district area list on rescue area service list

mikkoisa:
Units will be removed from rescue areas. Replace service unit list with area list (same as rescue districts and rescue sub districts).

Fix area focus error


mhieta (approved):
Looks good 👍

PR #730, opened 2021-09-09
Fix/map clicks

mikkoisa:
- New MapContainer component is immutable, so classNames cannot be updated. This is why measuring mode cursor changes needed to be moved to measuring component.
- Measuring mode had a bug that made the measuring markers lose interactive status after map zoom (cannot be clicked or dragged). Added interactive class manually after each zoom. This bug is present on current production version.
- Area markers used old click functionality that does not work anymore on newer react-leaflet version. Changed to new

### DIFF
--- a/src/views/MapView/components/DistanceMeasure/DistanceMeasure.js
+++ b/src/views/MapView/components/DistanceMeasure/DistanceMeasure.js
@@ -15,13 +15,23 @@ const DistanceMeasure = (props) => {
   } = global.rL;
 
   const theme = useTheme();
-
-  const [clickedPoint, setClickedPoint] = useState(null);
   const [icon, setIcon] = useState(null);
 
   useMapEvents({
     click(e) {
-      setClickedPoint(e.latlng);
+      setMarkerArray([...markerArray, e.latlng]);
+      setLineArray([...lineArray, e.latlng]);
+    },
+    zoom() {
+      // Distance markers lose interactive status after zooming. Add it back manually
+      setTimeout(() => {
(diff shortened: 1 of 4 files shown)


src/views/MapView/components/DistanceMeasure/DistanceMeasure.js
```diff
@@ -15,13 +15,23 @@ const DistanceMeasure = (props) => {
...
+      setMarkerArray([...markerArray, e.latlng]);
+      setLineArray([...lineArray, e.latlng]);
+    },
+    zoom() {
+      // Distance markers lose interactive status after zooming. Add it back manually
+      setTimeout(() => {
+        const measureMarkers = document.getElementsByClassName('leaflet-marker-draggable');
+        if (measureMarkers.length) {
+          measureMarkers.forEach((marker) => {
+            marker.classList.add('leaflet-interactive');
```
Dremars, line 31: Why does this lose leaflet-interactive class? Is this a bug on leaflet side?

Dremars (approved):

src/views/MapView/components/DistanceMeasure/DistanceMeasure.js
mikkoisa: I couldn't figure out what causes it. Other markers do not lose interactive class on zoom, so it must be caused by how ReactLeaflet handles this component.

Dremars: Ok. Eitherway this seems to fix it

PR #733, opened 2021-09-17
Feature/matomo spa update

Dremars:
New matomo SPA tracking based on history update using `@datapunt/matomo-tracker-js` library.

- Added configurations for accessibility dimensions: mobility and senses

Also refactoring SearchBar component to functional. Leftover from react tracking solution.


mikkoisa (approved):
Code changes look good!

PR #734, opened 2021-09-20
Feature/area view unit progress bar

mikkoisa:
Area view geographical unit fetch didn't have proper progress bar, which was not helpful for users.
- Added progress update option to concurrent search and
- Updated area view unit fetch redux structure to allow progress fetch

### DIFF
--- a/src/utils/newFetch/HTTPClient.js
+++ b/src/utils/newFetch/HTTPClient.js
@@ -26,7 +26,7 @@ export default class HttpClient {
 
   onError;
 
-  onNext;
+  onProgressUpdate;
 
   constructor(baseURL) {
     this.baseURL = baseURL;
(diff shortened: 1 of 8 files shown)


src/utils/newFetch/HTTPClient.js
```diff
@@ -133,7 +133,7 @@ export default class HttpClient {
     return this.fetch(endpoint, this.optionsToSearchParams(newOptions, true), 'count');
   }
 
-  getConcurrent = async (endpoint, options) => {
+  getConcurrent = async (endpoint, options, progressCallback) => {
```
Dremars, line 136: I wanted to avoid this kind of props drilling. Can we use current onNext functionality for this progressCallback? Possibly rename it to onProgressUpdate if onNext is too specific to next functionality in sequential fetching.

Dremars (changes requested):

src/utils/newFetch/HTTPClient.js
mikkoisa: Okay 👍 I was trying to be as careful as possible with this. OnNext wasn't working with concurrent search, so I will create onProgressUpdate to handle these.

mikkoisa: I renamed onNext to onProgressUpdate and used it here as you suggested, seems to work well!

Dremars (approved):

PR #750, opened 2021-11-09
Fix address bar enter click

mikkoisa:
Pressing enter on address bar with an address already selected cleared the address bar.


mhieta (approved):
👍

PR #752, opened 2021-11-10
Fix/search suggestion improvements

mikkoisa:
- Reduced amount of search suggestions to 5
- Disabled result ordering by match
- Improved area suggestions


mhieta (approved):
Ok 👍

PR #753, opened 2021-11-22
Feature/embed bbox option

mikkoisa:
New option for embedder tool to set embedded map bounds by moving the map preview.


mhieta (approved):
Good 👍

PR #768, opened 2022-01-13
Send searches with no results to matomo

mikkoisa:
- Added custom dimension to Matomo that logs search queries with no results


mhieta (approved):
LGTM

PR #781, opened 2022-01-31
Add option for multiple unit embedding

mikkoisa:
Adding parameter "units" with list of id:s to search page url will show only the selected units.
Example url: https://palvelukartta.hel.fi/fi/embed/search?units=8215,51342


mhieta (approved):
LGTM 👍

PR #783, opened 2022-02-09
Fix/unit price links

mikkoisa:
- Units with link price data (www data) did not render as links. Added missing link elements to unit price info
- Removed the text "events" from unit view third tab, since events have been moved to first tab


mhieta (approved):
LGTM 👍

PR #793, opened 2022-03-02
New search

mikkoisa:
The new search


Dremars:
While testing area-view's address seach functionality there as an issue with 400 bad requests. Seems like parameter `input` has been changed to `q`. This should be checked before release

PR #799, opened 2022-03-22
Embedded view unit list

mikkoisa:
- Added option to embedded view parameter to include list of units next to the embedded map. This improves embedded view accessibility.
-  Created a new hook that defines what data should be shown on map. Previously this was done as a complicated function inside MapView component. Now the units shown on map can be accessed by other components, such as this new embedded unit list.


mhieta (approved):
LGTM 👍

PR #808, opened 2022-05-02
Fix duplicate period services

mikkoisa:
Bug: school service list showed duplicate services. Example: https://palvelukartta.hel.fi/fi/unit/6926/educationServices
The extended list showed the services for both periods in the same list. 

Fixed this by adding period to url to define which period services should be shown (example url: unit/6926/educationServices/2022–2023)


mhieta (approved):
LGTM 👍

Dremars (approved):

PR #809, opened 2022-05-02
Feature: Use digitransit for route link if destination is outside of …

Dremars:
…specified municipalities

Use digitransit for route links for destinations that have municipality outside of specified HSL municipalities


Dremars:
Still need to actually check the link text since it has text referencing HSL and saying it's not accessible service. We need to check if we remove this text for digitransit links and can we tell it's not accessible service or not?

mikkoisa (approved):

PR #813, opened 2022-05-11
Update/libraries

Dremars:
- Update core-js, express, nodemon, turf, and eslint-plugin-react-hooks libraries
- Remove react-jss
- Move eslint-plugin-react-hooks and nodemon to devDependencies
- Change urijs to use ES6 imports


mikkoisa (approved):
Tested, looks good and works 👍

PR #814, opened 2022-05-16
Update/react to 17

Dremars:
Updated react to 17 and and added  react testing library. Fixed some of the tests to use new testing library but we should fix everything after material ui update


mikkoisa (approved):
LGTM 👍

PR #815, opened 2022-05-17
New embedder tool

mikkoisa:
New updated visual look for the embedder tool.

Missing translations


src/components/TopBar/TopBar.js
```diff
@@ -236,35 +239,41 @@ const TopBar = (props) => {
...
-            <DesktopComponent>
-              {!smallScreen ? (
-                <div className={classes.settingsButtonsContainer}>
-                  <Typography component="h2" variant="srOnly">
-                    <FormattedMessage id="settings" />
-                  </Typography>
-                  {renderSettingsButtons()}
-                </div>
-              ) : (
+            {simpleElement // SimpleElement prop removes top bar buttons
```
Dremars, line 242: Maybe this could be named hideButtons etc. to be more descriptive of what actually happens. Simple element doesn't tell exactly what this prop does

Dremars (changes requested):
Code seems mostly fine. I'll write extra comments related to visual issues

Dremars:
image
These scroll bars are making this quite hectic looking. Maybe those bottom scroll bars could be removed.

Also I found it confusing especially on right column that there is a big separation from the elements and vertical scrollbar. Maybe the padding could be reduced so the scrollable content is closer to the bar. This would improve association that the content is scrollable. 

It's also slightly confusing that the show list of units buttons are checkboxes on same level since both cannot be true at the same time. I feel those should be radio buttons left, bottom with extra none option by default

src/components/TopBar/TopBar.js
mikkoisa: Fixed!

Dremars (changes requested):
Seems like hide unit list radio option doesn't work if you've first selected to show lists and then press don't show.
image

Also do we want to show the option for showing list when there is less than 2 units shown? For example unit page or showing specific address etc?

Dremars (approved):
Looks OK

PR #816, opened 2022-05-17
Mobile embedded view fixes

mikkoisa:
On smaller embedded views (mobile) marker popups were hard to use. Servicemap logo would be on top of the popup and usually popup were opened outside of screen bounds.

Moved logo to bottom and added autoPan to marker popups

Also improved mobile click responsiveness


Dremars (approved):
Seems ok. Pan feature actually feels pretty good on small screens

PR #817, opened 2022-05-18
Update/material UI to v5

Dremars:
Update material UI to v5


Dremars:
Issues noticed in NVDA accessibility testing:
- After closing settings view focus is not moving back to settings button that oppened the view

mikkoisa:
Tested with Mac Voiceover, no problems found.

Search results orderer styles have changed though.
Näyttökuva 2022-5-19 kello 8 33 46

mikkoisa (approved):

PR #818, opened 2022-05-23
Fix/unit address handling

Dremars:
Changed getAddressFromUnit to create address string by adding each existing data to string and returning it
Made small improvement to Markercluster tooltip popup handling


mikkoisa (approved):
Looks good 👍

PR #821, opened 2022-05-31
Fix event button count bug

mikkoisa:
Event count broke when fix was made for educational service counts. Fixed this by adding different values for each case.


Dremars (commented):
Seems to work fine. Can you add a unitTest for this so we can find out if it breaks again?

mikkoisa:
Ok 👍

Dremars:
I think unit tests would be bettter for checking components inner functionality. Lets use react testing library for this. It also means we can strip ReactSelectors from use

Dremars (approved):

PR #822, opened 2022-06-02
Fix/browser tests

Dremars:
Fixed testcafe tests that got broken with react / materialUI updates. ReactSelector doesn't seem to work in most cases and it isn't really needed in most cases.


mikkoisa (approved):
Looks good 👍

PR #823, opened 2022-06-02
Change theme to follow v5 structure

Dremars:
adaptV4Theme was giving bunch of warnings which was causing issues with test result readability. Changed adapter to use

### DIFF
--- a/src/themes/index.js
+++ b/src/themes/index.js
@@ -1,5 +1,4 @@
 import { createTheme } from '@mui/material/styles';
-import { adaptV4Theme } from '@mui/material'
 
 const focusIndicator = {
   outline: '2px solid transparent',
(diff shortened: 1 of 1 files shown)


src/themes/index.js
```diff
@@ -268,29 +355,29 @@ export const paletteDark = {
 };
 
 // Themes
-const SMTheme = createTheme(adaptV4Theme({
+const SMTheme = createTheme({
   props,
-  overrides: overrides('default'),
```
mikkoisa, line 273: Seems like overrides object is not used anymore so it could be removed (starting on line 29).

mikkoisa (changes requested):

src/themes/index.js
Dremars: Fixed in e741fdb

mikkoisa (approved):

PR #825, opened 2022-06-07
Add better label for result order select element

mikkoisa:
Result orderer select element was missing linked label. Now it should read the label when navigating to the element. (Accessibility report page 32)


mikkoisa:
[(1) Pudotusvalikon nimeäminen](https://trello.com/c/J3siVd1x/1161-1-pudotusvalikon-nime%C3%A4minen)

Dremars (approved):
Works very well now! Maybe we could add a unit test but otherwise ok

PR #829, opened 2022-06-10
Fix: Change AddressSearchBar to use mui styled utility

Dremars:
Change AddressSearchBar to use new mui styled utility


src/components/AddressSearchBar/AddressSearchBar.js
```diff
@@ -196,8 +195,39 @@ const AddressSearchBar = ({
...
+  backgroundColor: '#fff',
+}));
+
+const StyledClear = styled(Clear)(() => ({
+  fontSize: 22,
+}));
+
+const StyledSearch = styled(Search)(({ theme }) => ({
+  color: 'rgba(0, 0, 0, 0.54)',
+  fontSize: 22,
```
mikkoisa, line 215: Looks good! We could use rem as font size here though, even if it is just for icon. So these 22px would be 1.375rem

Dremars: Good point

mikkoisa (approved):

PR #832, opened 2022-06-15
Fix area view embedder tool bug

mikkoisa:
Bug: Opening embedder tool in area view and then closing the tool caused a crash when returning to area view.

### DIFF
--- a/src/views/AreaView/AreaView.js
+++ b/src/views/AreaView/AreaView.js
@@ -1,3 +1,4 @@
+/* eslint-disable camelcase */
 import React, { useState, useEffect, useCallback } from 'react';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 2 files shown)


src/views/AreaView/AreaView.js
```diff
@@ -61,7 +62,7 @@ const AreaView = ({
   const searchParams = parseSearchParams(location.search);
   const selectedArea = searchParams.selected;
   // Get area parameter without year data
-  const selectedAreaType = selectedArea?.split(/([0-9]+)/)[0];
+  const selectedAreaType = selectedArea?.split(/([\d]+)/)[0];
```
Dremars, line 65: What does this do?

Dremars (approved):
This looks OK

PR #834, opened 2022-06-16
Fix/link underlines

mikkoisa:
Accessibility report pages 11-12: Some links did not have underline to indicate it is a link. 
Added link color to theme so that the color is more consistent.

### DIFF
--- a/src/themes/index.js
+++ b/src/themes/index.js
@@ -252,6 +252,9 @@ export const paletteDefault = {
     background: 'rgb(25, 100, 230)',
     border: '#fff',
   },
+  link: {
+    main: '#0000EE',
+  },
 };
 
 // Color palette for dark theme
@@ -293,6 +296,9 @@ export const paletteDark = {
     main: '#313131',
     background: '#A6A6A6',
     border: '#fff',
+  },
+  link: {
+    main: '#0000EE',
   },
 };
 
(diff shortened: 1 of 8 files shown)


src/themes/index.js
```diff
@@ -252,6 +252,9 @@ export const paletteDefault = {
     background: 'rgb(25, 100, 230)',
     border: '#fff',
   },
+  link: {
+    main: '#0000EE',
```
Dremars, line 256: Report had a point that we should have 3:1 contrast ration between normal text and links so we should adjust this slightly. #3333FF should be 3.05:1 compared to black text

src/themes/index.js
```diff
@@ -294,6 +297,9 @@ export const paletteDark = {
     background: '#A6A6A6',
     border: '#fff',
   },
+  link: {
+    main: '#0000EE',
```
Dremars, line 301: Report had a point that we should have 3:1 contrast ration between normal text and links so we should adjust this slightly. #3333FF should be 3.05:1 compared to black text

Dremars (changes requested):
It looks clearer now with underlines and having link color in contrast is also good change. Code seems OK. But I noticed that some links are missing link color:
- address view has a "tutustu alueisiin kartalla" link without color
- feedback view has "tietoja ja ohjeita..." link without color

Dremars (approved):

PR #835, opened 2022-06-16
Fix/large font styles

mikkoisa:
- Changing browser font size did not scale some texts that had font size as px. Changed these to use rem values.
- Fixed containers that would break with large font sizes
- Changed sone line height values from px to rem as well.


Dremars (changes requested):
Looking great! Noticed some issue with mobile view and large fonts. Topbar  map and menu buttons are on top of each other.
image

Dremars (approved):

PR #840, opened 2022-07-04
Fix/area view embed list

mikkoisa:
Bug fix. Check linked Trello card for details

Previously area units did not use useMapUnits hook to handle maker rendering. Embedded unit list uses this hook to define the list. Moved rendering logic from District component to useMapUntis hook to fix this.
  
Also fixed area view parking units bug.


mikkoisa:
[Bugi: Aluenäkymän upotuksessa voi valita näyttää toimipisteet mutta listauksessa ei tule mitään näkyviin](https://trello.com/c/M5rPyBPE/1180-bugi-aluen%C3%A4kym%C3%A4n-upotuksessa-voi-valita-n%C3%A4ytt%C3%A4%C3%A4-toimipisteet-mutta-listauksessa-ei-tule-mit%C3%A4%C3%A4n-n%C3%A4kyviin)

src/views/MapView/utils/useMapUnits.js
```diff
@@ -84,10 +84,13 @@ const useMapUnits = () => {
   const adminDistricts = useSelector(state => state.address.adminDistricts);
   const addressUnits = useSelector(state => state.address.units);
   const serviceUnits = useSelector(state => getServiceUnits(state));
-  const districtUnits = useSelector(state => getFilteredSubdistrictUnits(state));
+  const districtPrimaryUnits = useSelector(state => getDistrictPrimaryUnits(state));
+  const districtServiceUnits = useSelector(state => getFilteredSubdistrictUnits(state));
```
Dremars, line 94: What's the difference between these 2 unit sets? Do these have duplicated with each other since they are merged later to areaViewUnits variable?

mikkoisa: The selectors that are called here handle the unit sets so that there are no duplicates. First one is used on the first tab in Area view and the other in the second tab. That means only one of them has units and the other is empty, based on user's selected tab. 

I added a commit now to make the code a little bit clearer.

Dremars (approved):

PR #844, opened 2022-07-26
Optimize/hearing map to new fetch api

Dremars:
Change hearing map fetching to use our new fetching logic


mikkoisa (approved):
Looks good 👍

PR #845, opened 2022-08-04
Update/add unit view tests

Dremars:
Added browser tests for unit view for basic functionality
Added getLocation utility and changed tests to use same utility function


mikkoisa (approved):
Looks really good 👍

PR #850, opened 2022-08-10
Update settings browser tests to check each category is tested and fo…

Dremars:
…cus moves back to correct button


mikkoisa (approved):
Nice improvement!

PR #852, opened 2022-08-12
Prevent button focus showing on click

Dremars:
After MUI update button has shown keyboard focus indicator on click. This was caused by old focus style overrides being left in. I've removed the old focus styles so we can let material UI handle focus indicators with  Mui-focusVisible class


mikkoisa (approved):
I can't replicate this problem on Mac + Safari. However the fix looks good 👍

PR #858, opened 2022-08-29
Update/search tests

Dremars:
Update search browser and jest tests. Moved accessibility testing more to jest and added result orderer and pagination tests to search view.


mikkoisa (approved):
Looks good! Good idea to move providers to single file

PR #860, opened 2022-09-01
Add sub group contact informations to unit view

Dremars:
Add new kindergarten group contacts.  Data is part of connections with type "SUBGROUP" These were not shown previously but this PR introduces UI changes to show this data. Since subgroups are not necessarily kindergarten only data we've changed the text to be more generic in case someone adds subgroups for other services.


mikkoisa (approved):
LGTM

PR #861, opened 2022-09-02
Accessibility/navigation landmarks

Dremars:
Eficode report page 23

Added navigation landmarks for topbar language selection, settings and home view navigation buttons.


mikkoisa (approved):
Seems to work well 👍

PR #865, opened 2022-09-09
Fix/update broken axe and pagination tests

Dremars:
We've had broken test going through since we had TestCafe .only value slip to develop. This fixes broken pagination tests and navigation aria attribute which was breaking our Axe tests


mikkoisa (approved):
Looks good 👍

PR #871, opened 2022-09-26
Fix/area view zooming

mikkoisa:
In current version when an area is selected, the map zooms to address area but then zooms to fit all areas to map.
Now if user has selected an address, the map always zooms to that area only.


src/views/AreaView/AreaView.js
```diff
@@ -53,7 +53,7 @@ const AreaView = ({
   const dispatch = useDispatch();
   const location = useLocation();
   const history = useHistory();
-  const localAddressData = useSelector(state => state.districts.districtAddressData);
+  const addressDistrict = useSelector(state => getAddressDistrict(state));
```
Dremars, line 56: This is fine but these could be used without needing to pass state to getAddressDistrict function. `useSelector(getAddressDistrict)`

Dremars (commented):
I'm slightly confused as to what this does. I'm not able to see the difference to test version. Can you describe order of actions that should show the difference?

src/views/AreaView/AreaView.js
mikkoisa: I think getAddressDistrict needs the state passed, since it is a Reselect selector. Otherwise it will crash.

Dremars: It should still get state passed since the useSelector passes state to function it receives

Dremars (approved):

src/views/AreaView/AreaView.js
mikkoisa: I see! I didn't know it works that way! Much better like that 👍

PR #882, opened 2022-10-06
Fix/accessibility sorting

mikkoisa:
Fixed accessibility settings related bugs:
- Sorting search results by accessibility did not work correctly
- If user is on unit accessibility tab, sharing unit page with share button with accessibility settings did not form url correctly
- Fixed small things that caused errors in console


Dremars (approved):
Code looks good and some nice improvements. If these changes work well with screen readers then all OK

PR #891, opened 2022-10-11
Add opening hour object to unit data

mikkoisa:


mhieta (approved):
LGTM 👍

Dremars:
> Does the structure of the data affect required listing order? Should confirm this before merging

Seems we can just render them in any order so we can continue with this

PR #893, opened 2022-10-12
Feature/statistical geometry

Dremars:
**Statistical geometry feature**
Just like in the old servicemap we wanted to show statistical data on map. This feature implements that that functionality. We've decided that the best accessible way to show these statistical districts is by using same functionality as in AreaView. This way services can be browsed in same context with the actual statistical district data.

This implementation adds the following:
- New section into AreaView,
- Changes AreaView browsing to vertical accordions
- Allows selection of statistical districts and services
- Show statistical districts on map when statistical section is open
- Show statistical district value information on top right corner of the map


Dremars:
Missing translations

mikkoisa (commented):
Checked accessibility with Voiceover, no problems found except this heading level that should be level 3 instead of 5 I think. 
Näyttökuva 2022-10-31 kello 8 33 16

mikkoisa (approved):

PR #903, opened 2022-10-18
Feature/front page

Bansku:
Refactored homeview as part of UI refreshing


Bansku:
@mikkoisa add webpack loader to jpgs and webp please!

PR #914, opened 2022-11-09
Feature/simple pageview tracking

Dremars:
There was a need for simple pageview tracking only tracking count, embedded, and mobile info. Since Matomo can't be used without user permissions and embeds don't work well with cookiehub window popping every time on top of embedded view. This solution adds tracking that sends info to backend which counts pageviews anonymously.

Feature is hidden behind a feature flag FEATURE_SERVICEMAP_PAGE_TRACKING which is false by default.

This change will require environment variable changes


mikkoisa (approved):
Code looks good 👍 env.example file should be updated too to reflect these changes. Also I wonder if Turku uses similar version in API url?

PR #921, opened 2023-01-04
Improve Wellbeing areas support

mhieta:
Add a better support for wellbeing areas in Uusimaa-region.

This PR has two fixes:
- Remove unneeded city-part from contract text --> this is no longer needed as the contract text itself will have all the necessary information, including the city-part when needed
- Add Uusimaa-region municipalities to translations --> fix to overcome JS-errors with some new Uusimaa-region municipality


mikkoisa (approved):
Looks good 👍

PR #929, opened 2023-02-09
Check for mapPane existence before resolving center coordinates

vitalis0itu:
So our code failed on NPE because `this._mapPane` was `undefined` here https://github.com/Leaflet/Leaflet/blob/v1.7.1/src/map/Map.js#L1476 on atleast one instance of rerender after map type switch. Now the fix is to check for `mapPane` existence before resolving "center".

image

Could there be a more elegant solution by tweaking hooks? Probably yes.

Thoughts, questions and todo
- [x] `prevMap.props` seems to be always empty - Fixed to be `options` instead of `props`
- Why is this [src/redux/actions/map.js#](https://github.com/City-of-Helsinki/servicemap-ui/blob/develop/src/redux/actions/map.js#L4) `setMapRef` method used in `whenCreated` hook of leaflet map.
- [x] should fix browser tests - fixed

### DIFF
--- a/src/utils/dataVisualization.js
+++ b/src/utils/dataVisualization.js
@@ -30,7 +30,7 @@ class DataVisualization {
 
   POPULATION_FORECAST_STRING = '_population_forecast';
 
-  CURRENT_YEAR = new Date().getFullYear() - 1;
+  CURRENT_YEAR = new Date().getFullYear() - 2;
 
   FORECAST_YEAR = new Date().getFullYear() + 5;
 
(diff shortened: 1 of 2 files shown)


vitalis0itu:
browser tests can be fixed either by changind
```
-  CURRENT_YEAR = new Date().getFullYear() - 1;
+  CURRENT_YEAR = new Date().getFullYear() - 2;
```
here https://github.com/City-of-Helsinki/servicemap-ui/blob/develop/src/utils/dataVisualization.js#L33

Or tweaking output of `https://api.hel.fi/servicemap/v2/administrative_division/?page=1&page_size=500&geometry=true&type=statistical_district` to contain 2022 data and 2027 forecat

mhieta (approved):
LGTM 👍

PR #932, opened 2023-02-15
Fix unitView when unit does not have location or geometry prop

vitalis0itu:
https://trello.com/c/MBJENmmV


mhieta (approved):
Good :shipit:

PR #945, opened 2023-03-14
Use "this year" + 4 years for forecast as that is the only forecast d…

vitalis0itu:
…ata given

See changes in https://github.com/City-of-Helsinki/servicemap-ui/pull/929/files


mhieta (approved):
Korjaus Ok 👍

PR #961, opened 2023-03-31
Fix vantaa parking

vitalis0itu:


mhieta (approved):
Good 👍

mhieta (approved):
Looks better now

PR #1006, opened 2023-08-11
Remove with styles

vitalis0itu:


vitalis0itu:
Transit marker test is failing in `MapTest.js`

PR #1153, opened 2024-06-04
Change address location marker icon

japauliina:
Update the icon of the address location marker to similar as used in user location marker.

Old marker:
vanha

New marker:
uusi


mhieta (approved):
Great work! 💯

PR #1154, opened 2024-06-06
Fix: show parking units in embedder tool

japauliina:
Parking spaces are mostly area data, but there are few category exceptions that are units added to the area view:
- parking garages
- shared car parking
- accessible street parking

The issue was that these units were not shown in the embedder tool when selected from the area list. This fix adds parameters to the embedder tool to show these units when their category is selected.

Screenshot of a working scenario:
parking_spaces

[Refs](https://trello.com/c/d1B3dJqK/1560-vantaan-pys%C3%A4k%C3%B6intialueisiin-t%C3%A4pp%C3%A4-my%C3%B6s-upotukseen-mukaan)


mhieta (approved):
LGTM!

PR #1155, opened 2024-06-06
Fix: fit parking units to the map

japauliina:
Selecting category of parking units from area list did not zoom and fit the map as in area data. Fix this by using `fitUnitsToMap` function when category with parking units is selected.


mhieta (approved):
Looks good! :shipit:

PR #1157, opened 2024-06-10
Do not display unit source if unit is private service

japauliina:
No need to display unit source if unit type is private service.

[Refs](https://trello.com/c/UZfRKrit/1359-palvelukartalla-n%C3%A4ytett%C3%A4v%C3%A4t-tiedot-korjaus)


mhieta (approved):
Looks good!

PR #1161, opened 2024-06-13
Remove settings from the home view

japauliina:
As the setting can be found also in the navigation bar, remove them from the home view to give the news etc. more space.

Previously:
old

Now:
new

[Refs](https://trello.com/c/DauOldpG/1571-poistetaan-toinen-omat-asetukset-kohta)


mhieta (approved):
LGTM!

PR #1162, opened 2024-06-17
Fix zooming when selecting area from list

japauliina:
When selecting a neighborhood, postcode or major district from the area list, the map should zoom to the selected area. Fix issue where the map did zoom but only momentarily and then zoomed back to the original view.

zoom

[Refs](https://trello.com/c/uEDYl6iL/1578-bugi-kun-valitsee-postinumeron-listasta-n%C3%A4kym%C3%A4-zoomaa-vain-hetkellisesti)


mhieta (approved):
Good 👍

PR #1163, opened 2024-06-17
Fix: zoom to area when selected on map

japauliina:
When an area is clicked on the map focus and zoom to that area.

[Refs](https://trello.com/c/lId9fUvR/1214-1-bugi-kun-alueen-valitsee-aluen%C3%A4kym%C3%A4ss%C3%A4-n%C3%A4kym%C3%A4-ei-zoomaa-alueelle)


mhieta (approved):
LGTM

PR #1170, opened 2024-08-16
Fix navigating from embedder view to info and feedback view

japauliina:
Navigating from the embedder view to the info or feedback view did not work as expected. Instead of navigating to the correct view, the user was redirected the embed view.

[Refs](https://trello.com/c/3IP94sge/1442-1-upotusty%C3%B6kalusta-siirtyminen-ei-toimi-oikein)


mhieta (approved):
LGTM!

PR #1171, opened 2024-08-16
Save contrast theme to local storage

japauliina:
Save the contrast theme to local storage to keep the theme when refreshing the page or opening the site again.

[Refs](https://trello.com/c/EKSqeLor/1452-2-kontrastiteema-tumma-teema-ei-lataudu-refreshill%C3%A4)


mhieta (approved):
LGTM!

PR #1172, opened 2024-08-19
Make AddressSearchBar more accessible

japauliina:
Previously used aria-label was not descriptive enough. Use label with the title instead.

[Refs](https://trello.com/c/C1qbbfYr/1281-1-osoitehakukent%C3%A4n-aria-labelin-selvitys)


mhieta (approved):
LGTM!

PR #1183, opened 2024-09-04
Explicitly disable search engine selection screen

mhieta:
Explicitly disable search engine selection screen for browsertests.


mhieta:
Closing this as this is fixed in https://github.com/City-of-Helsinki/servicemap-ui/pull/1182

PR #1186, opened 2024-09-25
Fix: Handle all areas of MultiPolygon in UnitGeometry

japauliina:
The issue was that only the first area of a MultiPolygon was used in the UnitGeometry, meaning that only part of some areas were shown in the map. Updated the MultiPolygon case to iterate over all polygons.

Before the fix:
before

And after:
after

[Refs](https://trello.com/c/FpimYvAi/1553-luonnonsuojelualueiden-alue-geometrian-p%C3%A4ivitys-toimipisteisiin)


mhieta (approved):
Good!

PR #1187, opened 2024-10-04
Add Plainmap as new background map

mhieta:
plainmap

Add Plainmap as new background map.

Details:
- Add map configuration for plainmap
- Update layer orders in layer selector
- Update fi/sv/en-translations for 3D-maps


japauliina:
I would move the "Update fi/sv/en-translations for 3D-maps" change to a separate commit but otherwise looks good! Approving once the tests pass.

mhieta:
> I would move the "Update fi/sv/en-translations for 3D-maps" change to a separate commit but otherwise looks good! Approving once the tests pass.

Moved to a separate commit now.

japauliina (approved):
Nice work!

PR #1190, opened 2024-10-16
Fix module not found error

japauliina:
Add missing http-status-typed dependency to resolve module not found error during build process.


mhieta (approved):
LGTM!

PR #1191, opened 2024-10-16
Heightprofile support for 3D-lines

mhieta:
Add ElevationControl to MapView.

Desktop view:
heightprofile

Mobile view with interactions:
mobile view with interactions

Mobile view without interactions:
mobile view without interactions

Breakdown:
- Add Leaflet.Heightgraph and D3 to project
- Automatically display heightprofile in unit-page when unit has 3D linestring geometry in it
- Allow heightprofile closing via heightprofile close button
- Allow heightprofile re-opening from the map via new area-chart button
- Add and finetune heightprofile control styles
- Add mobile layouts for heightprofile control
- Update tests
- Localize to fi/sv/en


src/views/MapView/components/ElevationControl/ElevationControl.js
```diff
@@ -0,0 +1,129 @@
...
+
+const ElevationControl = ({ data, isMobile}) => {
+  const intl = useIntl();
+  const map = useMap();
+  const currentPage = useSelector(getPage);
+  const [geometryData, setGeometryData] = useState(null);
+
+  useEffect(() => {
+    const getUnitGeometry = (unit) => {
+      if ((currentPage === 'unit')) {
```
japauliina, line 15: You could use early returns here to avoid nesting:
```
if (currentPage !== 'unit') {
    return null;
  } 
```
..etc. But it is up to you to decide.

src/views/MapView/components/ElevationControl/ElevationControl.js
```diff
@@ -0,0 +1,129 @@
...
+    return () => {
+        map.removeControl(control);
+        map.removeLayer(displayGroup);
+    }
+  });
+
+  return null;
+}
+
+export default ElevationControl;
```
japauliina, line 129: Missing last line.

japauliina (commented):
Very nice work! ✨  Left some minor comments but otherwise looks good!

src/views/MapView/components/ElevationControl/ElevationControl.js
mhieta: Updated.

src/views/MapView/components/ElevationControl/ElevationControl.js
mhieta: Added.

japauliina (approved):

PR #1192, opened 2024-10-18
Migrate to React v18

mhieta:
Migrate to React v18.
Also fix various JavaScript issues and warnings.

Details:
- Upgrade packages:
  - React and React DOM 17.0.0 --> 18.3.1
  - React Helmet 5.2.1 --> 6.1.0
  - React-Leaflet 3.2.1 --> 4.2.1
  - Leaflet 1.7.1 --> 1.9.4
  - Leaflet Markercluster 1.4.1 --> 1.5.3
  - Material UI v5 --> v6
  - HDS React 3.6.0 --> 3.10.1
  - Sentry v5 --> v8
  - Testcafe React Selectors 4.1.5 --> 5.0.3
- Update to use ReactDOM hydrateRoot
- Update TestCafe configuration and browser tests
- Update Jest snaphots
- Change defaultProps to JS default parameters
- Update DefaultLayout CSS
- Update to use ListItemButton instead of ListItem
- Update MapComponent whenCreated property to ref
- Upgrade District tooltips for Leaflet v.1.9.4
- React Helmet v6 update will remedy deprecated React component lifecycle warnings
- Set default values for unit names
  - ReadFeedbackLink requires fi/sv/en unit names, thus introduce defaults for them
- Change nested link-tag to division
  - This fixes warning: `validateDOMNesting(...): <a> cannot appear as a descendant of <a>`
- Render DescriptionText SSR and CSR the same way
  - This fixes warning: `validateDOMNesting(...): <div> cannot appear as a descendant of <div>`
- Disable reservations fetching for now
  - Reservations API is outdated, thus disable this integration for now

PR #1195, opened 2024-10-31
Revert "Migrate to React v18"

japauliina:
Revert City-of-Helsinki/servicemap-ui#1192 temporarily until issues with the build are resolved.


mhieta (approved):
Good. We need to do this because of dependent CI-issues.

PR #1197, opened 2024-11-11
Fix visually impaired in ACCESSIBILITY_MAPPING

japauliina:
The accessibility shortcomings were not shown for visually impaired as the mapping had wrong naming for it.


mhieta (approved):
LGTM!

PR #1204, opened 2024-11-19
fix: bottomNav settings add close button PL-18

mikkojamG:
Add close button to "My settings" and "Map tools" settings widgets, so that they can be closed easily when navigating on keyboard.

Screenshot 2024-11-19 at 14 00 02


mhieta (approved):
Good!

PR #1206, opened 2024-11-20
Fixing  a11y status messages PL-17

mikkojamG:
- settingsDropdown should announce clearing selections
- serviceTree, checkbox should announce "mixed" state, HDS Checkboxes do this internally
- Clicking on additional entrances link should focus on tab


japauliina (approved):
I would prefer to see the stylistic changes in separate commits but otherwise looks good.

PR #1208, opened 2024-11-26
Migrate to React v18 (vol 2)

japauliina:
- Re-apply changes of [#1192](https://github.com/City-of-Helsinki/servicemap-ui/pull/1192)
- Temporary remove HDS and comment out code that uses it
- Use legacy-peer-deps flag in install


mhieta (approved):
Looks good! :shipit:

PR #1210, opened 2024-11-27
Revert "Temporary remove HDS and comment out code that uses it"

japauliina:
Add HDS back now that the React v18 upgrade issues are fixed.


mhieta (approved):
LGTM! 💯

PR #1212, opened 2024-11-29
Improve fetchUnits to increase search radius when necessary

japauliina:
By default, fetch units within 100 meters but if there are less than 50 units, fetch units within 500 meters. 50 units is the maximum amount that is currently wanted to be shown in the map.

[Refs](https://trello.com/c/AHY9UnRZ/1619-haun-datan-optimointi)


mhieta (approved):
Excellent! 💯

PR #1217, opened 2025-01-15
Allow unit MultiLineString scrolling

mhieta:
Allow unit MultiLineString scrolling when unit geometry has multiple line segments in it.
Browsing segments happen through new next- and previous-buttons in the height profile component's left bottom corner.

Also added TestCafe browser tests for the height profile.

Desktop-layout:
servicemap-height-profile

Mobile-layout:
servicemap-height-profile-mobile

Refs: [PL-41](https://helsinkisolutionoffice.atlassian.net/browse/PL-41)

[PL-41]: https://helsinkisolutionoffice.atlassian.net/browse/PL-41?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/views/MapView/components/ElevationControl/ElevationControl.js
+++ b/src/views/MapView/components/ElevationControl/ElevationControl.js
@@ -1,91 +1,101 @@
 import { useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
 import { useIntl } from 'react-intl';
 import { useSelector } from 'react-redux';
 import { useMap } from 'react-leaflet';
 import { getPage } from '../../../../redux/selectors/user';
 
-const ElevationControl = ({ data, isMobile}) => {
+const ElevationControl = ({ unit, isMobile}) => {
   const intl = useIntl();
   const map = useMap();
   const currentPage = useSelector(getPage);
-  const [geometryData, setGeometryData] = useState(null);
+  const [geometryIndex, setGeometryIndex] = useState(0);
+  const [geometry, setGeometry] = useState(null);
 
-  useEffect(() => {
-    const getUnitGeometry = (unit) => {
-      if (currentPage !== 'unit') {
-        return null;
-      }
-      const { geometry_3d } = unit;
-      if (geometry_3d) {
-        const coordinates = geometry_3d?.coordinates;
-        if (coordinates && geometry_3d.type === 'MultiLineString') {
-          return {
-            ...geometry_3d,
-            coordinates: coordinates[0],
-          };
-        }
-      }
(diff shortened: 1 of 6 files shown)


src/views/MapView/components/ElevationControl/ElevationControl.js
```diff
@@ -1,91 +1,98 @@
...
           };
         }
       }
       return null;
     };
-
-    setGeometryData(getUnitGeometry(data));
-  }, [data]);
+    setGeometry(getUnitGeometry(unit));
+  }, [unit, geometryIndex]);
```
mikkojamG, line 33: I think currentPage should be included as a dependency here 🤔

src/views/MapView/components/ElevationControl/ElevationControl.js
```diff
@@ -1,91 +1,98 @@
...
   const intl = useIntl();
   const map = useMap();
   const currentPage = useSelector(getPage);
-  const [geometryData, setGeometryData] = useState(null);
+  const [geometryIndex, setGeometryIndex] = useState(0);
+  const [geometry, setGeometry] = useState(null);
 
   useEffect(() => {
-    const getUnitGeometry = (unit) => {
+    const getUnitGeometry = (unitData) => {
```
mikkojamG, line 16: This function could me moved outside of useEffect.

src/views/MapView/components/ElevationControl/ElevationControl.js
```diff
@@ -1,91 +1,98 @@
...
       }
       return null;
     };
-
-    setGeometryData(getUnitGeometry(data));
-  }, [data]);
+    setGeometry(getUnitGeometry(unit));
+  }, [unit, geometryIndex]);
 
   useEffect(() => {
```
mikkojamG, line 35: There is quite a lot of code inside this useEffect. I think it could be evaluated if everything needs to be here. 
Hook is missing dependency list at the end, this can potentially cause infinite updates. Empty list can be used if the hook needs to be run only once.

src/views/MapView/components/ElevationControl/ElevationControl.js
mhieta: Updated

src/views/MapView/components/ElevationControl/ElevationControl.js
mhieta: Made "best effort"-refactoring here for now.

mikkojamG (approved):

PR #1251, opened 2025-06-30
chore: migrate from npm to yarn PL-92

Riippi:
Change build system from npm to yarn.

https://helsinkisolutionoffice.atlassian.net/browse/PL-92

Starting with yarn without old lock file updated a bunch of dependencies. Because of that one unit test snapshot failed so I just updated it. It was about one material ui focus css class that appeared with new build. The button of that test seems to work ok on browser so it's probably something to do with test environment.


README.md
```diff
@@ -13,41 +13,41 @@ For development:
...
 
 Start dev server and runs the app in the development mode.<br>
 
-### `npm run build`
+### `yarn build`
 
 Builds the app for production to the `dist` folder.<br>
 
-### `npm run testBuild`
+### `yarn testBuild`
```
arpargo, line 24: should we format this to same format as usual commands in our projects? I'm thinking something like yarn build:dev/test.

PR #1345, opened 2026-04-24
fix: apifetcherrors sentry spam part 2  PL-239

Riippi:
Another PR for fixing apifetcherrors.

https://helsinkisolutionoffice.atlassian.net/browse/PL-239

Refs: PL-239


mikkojamG (approved):
👍 🦭

